PR #104, opened 2022-05-25
NAS-3743 added user service

hkad98:
- added user service with auxiliary files and tests


gdgate:

gdgate:

gooddata-sdk/gooddata_sdk/catalog/user/service.py
```diff
@@ -0,0 +1,97 @@
...
+    def get_user(self, user_id: str) -> CatalogUser:
+        user_dict = self._entities_api.get_entity_users(id=user_id, include=["userGroups"]).data.to_dict(
+            camel_case=False
+        )
+        return CatalogUser.from_dict(user_dict, camel_case=False)
+
+    def create_user(
+        self, user_id: str, authentication_id: Optional[str] = None, user_groups: Optional[list[str]] = None
+    ) -> None:
+        user_document = CatalogUserDocument.create_user(user_id, authentication_id, user_groups)
```
jaceksan, line 44: This new concept of Document classes and create_x methods is confusing me.
Why is it necessary only in the case of users and user groups?

gooddata-sdk/gooddata_sdk/catalog/user/service.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class CatalogUserService(CatalogServiceBase):
+    def list_users(self) -> List[CatalogUser]:
+        get_users = functools.partial(
+            self._entities_api.get_all_entities_users,
+            include=["userGroups"],
+            _check_return_type=False,
+        )
+        users = load_all_entities_dict(get_users, camel_case=False)
```
jaceksan, line 23: Why is load_all_entities_dict needed in the case of users?

jaceksan (changes requested):

gooddata-sdk/gooddata_sdk/catalog/user/service.py
hkad98: It is because of the issue we talked before. In case of users the translation from camel case to snake case is not performed by client class. In case of user groups there are currently no keys that would be change because of conversion, so the old approach can be used.

gooddata-sdk/gooddata_sdk/catalog/user/service.py
hkad98: I based that on apidocs where POST /api/entities/users uses object JsonApiUserInDocument and POST /api/entities/userGroups uses object JsonApiUserGroupInDocument. These objects are basically wrappers for  
JsonApiUserIn, JsonApiUserGroupIn. The reason why you might  find this confusing is because these are the first entity objects that are implemented using data classes.

jaceksan (approved):

yenkins:
### Sonar scan result


More detail, see in https://sonarqube-gate.intgdc.com/dashboard?id=gooddata-python-sdk-gate-PR104

To scan for vulnerabilities in dependencies and run unit tests (to get coverage report in sonar) please comment your PR with 'extended check sonar'.

gdgate:

- https://checklist.intgdc.com/job/compliance/job/gooddata-python-sdk-sonarqube-scan-zuul-gate/96/ : SUCCESS in 1m 13s
- https://checklist.intgdc.com/job/compliance/job/gooddata-python-sdk-fossa-scanning-zuul-el7/96/ : SUCCESS in 2m 56s (non-voting)
- https://checklist.intgdc.com/job/compliance/job/gooddata-python-sdk-merge-compliance-check/96/ : SUCCESS in 18s

PR #134, opened 2022-08-17
NAS-4254 manage dependent entities graph

hkad98:


gdgate:

gooddata-sdk/gooddata_sdk/catalog/base.py
```diff
@@ -65,3 +70,7 @@ def client_class() -> Any:
     def to_api(self) -> Any:
         dictionary = self._get_snake_dict()
         return self.client_class().from_dict(dictionary, camel_case=False)
+
+    @staticmethod
+    def _custom_converter() -> Optional[cattr.GenConverter]:
```
pcerny, line 75: Make custom converted result singleton. Current implementation creates as new converter for each `from_api` call but it is a static object.

gooddata-sdk/gooddata_sdk/catalog/base.py
```diff
@@ -65,3 +70,7 @@ def client_class() -> Any:
     def to_api(self) -> Any:
         dictionary = self._get_snake_dict()
         return self.client_class().from_dict(dictionary, camel_case=False)
```
pcerny, line 72: Not part of this PR, but `client_class` should return Type which is bound to Base class.

Could be `def client_class() -> Any:` replaced with `def client_class() -> Type[T]:` ?

gooddata-sdk/gooddata_sdk/catalog/base.py
```diff
@@ -26,7 +27,11 @@ def from_api(cls: Type[T], entity: Dict[str, Any]) -> T:
...
         """
-        return structure(entity, cls)
+        convertor = cls._custom_converter()
+        if convertor is not None:
+            return convertor.structure(entity, cls)
+        else:
+            return structure(entity, cls)
 
     @classmethod
     def from_dict(cls: Type[T], data: Dict[str, Any], camel_case: bool = True) -> T:
```
pcerny, line 37: Should not be custom_converter utilized also in `from_dict` method?

gooddata-sdk/gooddata_sdk/catalog/base.py
```diff
@@ -65,3 +70,7 @@ def client_class() -> Any:
     def to_api(self) -> Any:
         dictionary = self._get_snake_dict()
```
pcerny, line 71: How does attrs map `from_` to `_from`? If it works out of box, please document it somewhere.

gooddata-sdk/gooddata_sdk/catalog/workspace/entity_model/graph_objects/graph.py
```diff
@@ -0,0 +1,95 @@
...
+@attr.s(auto_attribs=True, kw_only=True)
+class CatalogDependentEntitiesResponse(Base):
+    graph: CatalogDependentEntitiesGraph
+
+    @staticmethod
+    def client_class() -> Type[DependentEntitiesResponse]:
+        return DependentEntitiesResponse
+
+    @staticmethod
+    def _custom_converter() -> Optional[cattr.GenConverter]:
```
pcerny, line 37: This looks like converter is registered only for `CatalogDependentEntitiesResponse` but I could ask for conversion of nested structure, like `CatalogDependentEntitiesGraph` or `CatalogDependentEntitiesEdge`.

Ideally the implementation should:
1. go through attribute metadata and build `hooks`
2. register hooks from all its attributes to converter instance

Step (1)  must be generic, step (2) ideally too (requires typing introspection). For simplicity, step(2) can be done manually in case nested structures are in use.

pcerny (changes requested):

gooddata-sdk/gooddata_sdk/catalog/base.py
hkad98: I do not think it should. Client class returns the related class of the client, which is unrelated to the Base class.

gooddata-sdk/gooddata_sdk/catalog/base.py
hkad98: That is a fair point. Let's discuss what we expect the dictionary to be because the client class works with `_from.` Should the dictionary also work with `_from`, or do we expect it to work with `from`? IMHO it could work with both, but then there will be an issue when we want to utilize `to_dict`. What then?

gooddata-sdk/gooddata_sdk/catalog/base.py
hkad98: There is no endpoint to edit the dependent entities graph, so the method `to_api` is never called. For that, reason I did not modify `to_api`. When the `to_api` method will be called on the dependent entities graph  it will raise an exception in the current setup. I propose to override this method for the dependent entities graph and return `NotImplemented`.

gooddata-sdk/gooddata_sdk/catalog/workspace/entity_model/graph_objects/graph.py
hkad98: That is a fair point. The issue could be resolved using the singleton you proposed in the previous comment. I will make a draft.

gooddata-sdk/gooddata_sdk/catalog/base.py
hkad98: It actually returns Type[ModelNormal], but there is an issue that there is a ModelNormal for each client.

gdgate:

pcerny (approved):

yenkins:
### Sonar scan result


More detail, see in https://sonarqube-gate.intgdc.com/dashboard?id=gooddata-python-sdk-gate-PR134

To scan for vulnerabilities in dependencies and run unit tests (to get coverage report in sonar) please comment your PR with 'extended check sonar'.

gdgate:

- https://checklist.intgdc.com/job/compliance/job/gooddata-python-sdk-sonarqube-scan-zuul-gate/123/ : SUCCESS in 56s
- https://checklist.intgdc.com/job/compliance/job/gooddata-python-sdk-fossa-scanning-zuul-el7/123/ : FAILURE in 13s (non-voting)
- https://checklist.intgdc.com/job/compliance/job/gooddata-python-sdk-merge-compliance-check/125/ : SUCCESS in 19s

PR #575, opened 2024-03-04
CQ-229: Add dimension sorting support to exec def assembly logic

Tjev:


gdgate:

gdgate:

gdgate:

gdgate:

gdgate:

gdgate:

gooddata-sdk/gooddata_sdk/visualization.py
```diff
@@ -405,6 +460,54 @@ def __repr__(self) -> str:
         return repr(self._filter)
 
 
+class VisualizationSortLocator:
+    def __init__(self, locator: dict[str, dict[str, str]]) -> None:
+        locator_key = list(locator.keys())[0]
```
hkad98, line 465: It is not safe to pass the dictionary here and take the first value of keys. The order in a dictionary can change. I suggest passing the locator and the type directly. There is no need to pass the whole dictionary. It is not stored anywhere, only accessed.

gooddata-sdk/gooddata_sdk/compute/model/execution.py
```diff
@@ -86,10 +98,63 @@ def is_one_dim(self) -> bool:
...
+
+    def _create_sort_keys(self, sort_keys: list[dict]) -> list[models.SortKey]:
+        api_sort_keys: list[models.SortKey] = []
+        for sort_key in sort_keys:
+            if "value" == list(sort_key.keys())[0]:
+                api_sort_keys.append(self._create_value_sort_key(sort_key))
+            elif "attribute" == list(sort_key.keys())[0]:
+                api_sort_keys.append(self._create_attribute_sort_key(sort_key))
+            else:
+                continue
```
hkad98, line 137: Is order really important in `sort_key.keys()`? If not, please do the following:

```python
value = sort_key.get("value")
attribute = sort_key.get("attribute")
if value is not None:
    api_sort_keys.append(self._create_value_sort_key(sort_key))
if attribute is not None:
    api_sort_keys.append(self. _create_attribute_sort_key(sort_key))
```

gooddata-sdk/gooddata_sdk/table.py
```diff
@@ -219,8 +230,12 @@ def _prepare_tabular_definition(
...
 ) -> ExecutionDefinition:
     dims = [
-        [a.local_id for a in attributes] if len(attributes) else None,
-        ["measureGroup"] if len(metrics) else None,
+        ExecTableDimension(
+            item_ids=[a.local_id for a in attributes] if attributes else None,
+        ),
+        ExecTableDimension(
+            item_ids=[_MEASURE_GROUP_IDENTIFIER] if len(metrics) else None,
+        ),
```
hkad98, line 238: Just out of curiosity. Why do conditions differ?

gooddata-sdk/gooddata_sdk/visualization.py
```diff
@@ -405,6 +460,54 @@ def __repr__(self) -> str:
...
+    def direction(self) -> SortDirection:
+        return _DIRECTION_STR_TO_SORT_DIRECTION[self._sort["direction"]]
+
+    @property
+    def attribute_identifier(self) -> str:
+        return self._sort["attributeIdentifier"] if self.type == SortType.ATTRIBUTE else ""
+
+    @property
+    def aggregation(self) -> Optional[str]:
+        return self._sort.get("aggregation")
```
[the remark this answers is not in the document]
Tjev: I'd assume the optional would still be needed as dict.get() returns None by default?

gooddata-sdk/gooddata_sdk/table.py
Tjev: Mistake on my side, they should be aligned :)

gooddata-sdk/gooddata_sdk/visualization.py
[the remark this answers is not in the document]
Tjev: Aggregation is often not included in the insight sort payload, so I would prefer to keep it optional if you agree.
The attribute_sort_type being conditioned on the presence of aggregation is something I decided to "copy" from the gooddata-ui-sdk.

I am not entirely sure why this is the correct way, but I considered the ui sdk to be the more complete exec-def assembly logic 😕.

gdgate:

gdgate:

gooddata-sdk/gooddata_sdk/table.py
```diff
@@ -258,7 +278,188 @@ def _create_dimension(bucket: VisualizationBucket, measures_item_identifier: Opt
...
+    @param sorts - sort items defined by SDK user
+    """
+    if not sorts:
+        return dims
+
+    non_measure_dim_idx = _get_dim_idx_for_predicate(dims, lambda x: _MEASURE_GROUP_IDENTIFIER not in x.item_ids)
+    measure_dim_idx = _get_dim_idx_for_predicate(dims, lambda x: _MEASURE_GROUP_IDENTIFIER in x.item_ids)
+    measure_dim = dims[measure_dim_idx] if measure_dim_idx else None
+
+    sorting: list[list[SortKey]] = [[] for _ in dims]
```
hkad98, line 398: `sorting: list[list[SortKey]] = [[] * len(dims)]`

gooddata-sdk/gooddata_sdk/visualization.py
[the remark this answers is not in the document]
hkad98: Sorry, I made a typo. I meant. 
```python
    @property
    def aggregation(self) -> str:
        return self._sort.get("aggregation", AttributeSortType.DEFAULT)
```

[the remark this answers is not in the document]
Tjev: I believe I cannot use the AttributeSortType here as aggregation contains values like "sum".
https://github.com/gooddata/gooddata-ui-sdk/blob/master/libs/sdk-model/src/execution/base/sort.ts#L64-L70
https://github.com/gooddata/gooddata-ui-sdk/blob/master/libs/sdk-model/src/execution/base/sort.ts#L189-L196

gdgate:

hkad98 (approved):

gooddata-sdk/gooddata_sdk/table.py
Tjev: Hmm, seems like this produces just a list with a single list for any value of len(dims)>1. We need list of multiple empty lists here (in case of len(dims)>1). So I would keep it as is if you agree.

gdgate:

gdgate:

gdgate:

gdgate:

gdgate:

- https://checklist.intgdc.com/job/compliance/job/gooddata-python-sdk-fossa-scanning-zuul/331/ : FAILURE in 1m 39s (non-voting)
- https://checklist.intgdc.com/job/compliance/job/gooddata-python-sdk-sonarqube-scan-zuul-gate/481/ : SUCCESS in 1m 06s
- https://checklist.intgdc.com/job/compliance/job/gooddata-python-sdk-merge-compliance-check/485/ : SUCCESS in 10s

PR #641, opened 2024-04-17
Update documentation pipelines - local preview and only 4 versions

Mara3l:


.github/actions/hugo-build-versioned-action/action.yaml
```diff
@@ -35,17 +36,34 @@ runs:
...
         chmod +x ./generate.sh
         ./generate.sh ${{ inputs.fetch-from }} master ${{ inputs.keep-master}}
-    - name: Hugo Build
-      uses: gooddata/gooddata-python-sdk/.github/actions/hugo-build-action@master
-      with:
-        hugo-env: ${{ inputs.hugo-env }}
-        working-directory: ${{ inputs.working-directory }}
-        base-url: ${{ inputs.base-url }}
+        npm install
+        hugo version
```
hkad98, line 68: What is the purpose of this command?

Mara3l: It is just for having better output, we want to be sure we know which version we are using in case of debugging. I know we state it in the action itself, but I would like to be extra sure. This also shows any minors, that might apply.

.github/actions/hugo-build-versioned-action/action.yaml
```diff
@@ -35,17 +36,34 @@ runs:
...
+      uses: peaceiris/actions-hugo@v2
+      with:
+        extended: true
+        hugo-version: '0.110.0'
+    - name: "Build"
+      env:
+        KEEP_MASTER: ${{ inputs.keep-master }}
+        THIS_BRANCH: ${{ inputs.this-branch }}
+        BASE_URL: ${{ inputs.base-url }}
+        HUGO_ENV: production
```
hkad98, line 59: Please, hide this in variable as well. I can imagine that we might want to use this action not for production.

.github/actions/hugo-build-versioned-action/action.yaml
hkad98: I think it is not necessary. The version is decided based on
````   
 - name: "Setup Hugo"
      uses: peaceiris/actions-hugo@v2
      with:
        extended: true
        hugo-version: '0.110.0'
```

.github/actions/hugo-build-versioned-action/action.yaml
```diff
@@ -35,17 +36,34 @@ runs:
...
+    - uses: actions/setup-go@v4
+      with:
+        go-version: '>=1.20.1'
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+    - name: Install dependencies
+      working-directory: ${{ inputs.working-directory }}
+      run: npm ci
+      shell: bash
```
hkad98, line 48: Is this necessary? If I am not mistaken this should be resolved by peaceiris/actions-hugo@v2 action.

hkad98: Or more precisely, based on the documentation node setup should follow after peaceiris/actions-hugo@v2

.github/actions/hugo-build-versioned-action/action.yaml
Mara3l: I know it is there, but it does mitigate a problem with debugging the logs. It's in the yaml itself, but this saves the extra step for someone who is not familiar with this workflow.

.github/actions/hugo-build-versioned-action/action.yaml
```diff
@@ -35,17 +36,34 @@ runs:
...
+    - name: "Setup Hugo"
+      uses: peaceiris/actions-hugo@v2
```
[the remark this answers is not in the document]
Mara3l: OK

.github/actions/hugo-build-versioned-action/action.yaml
hkad98: I am still not convinced. It is still redundant. The version can be found in the log from peaceiris/actions-hugo@v2

.github/actions/hugo-build-versioned-action/action.yaml
Mara3l: We can try to omit this, but I copied it from our main documentation and wanted to be extra sure, that it would run.

scripts/generate.sh
```diff
@@ -42,8 +44,24 @@ fi
...
 git fetch "$remote_name"
 
+latest_branches=()
+# Get all relevant rel/* branches, sort them, and pick the latest num_versions
+while IFS= read -r vers; do
+    latest_branches+=("$remote_name/rel/$vers")
+done < <(git branch -rl "$remote_name/rel/*" | sed 's|.*/rel/||' | sort -t. -k1,1n -k2,2n | tail -n"$num_versions")
+
+# Add special branches to the array (only rel/dev and master for now)
+special_branches=("$remote_name/rel/dev" "$remote_name/master")
```
hkad98, line 54: Why do we need master?

hkad98 (approved):

PR #658, opened 2024-04-22
DP-2808 - 404 message for older versions

quanb-duy:


hkad98:
Please have a look at the commits. One commit seems to be verified, but the second one does not.

Mara3l (approved):

PR #678, opened 2024-05-13
feat: add support for MotherDuck data source

jaceksan:


Makefile
```diff
@@ -5,7 +5,10 @@ NO_CLIENT_GD_PROJECTS_ABS = $(filter-out %client, $(wildcard $(CURDIR)/*gooddata
 NO_CLIENT_GD_PROJECTS_DIRS = $(foreach dir, $(NO_CLIENT_GD_PROJECTS_ABS), $(notdir $(dir)))
 # TODO: replace API_VERSION in the future by call to API
 API_VERSION="v1"
+# Generate from localhost
 BASE_URL="http://localhost:3000"
+# Generate from PROD
+# BASE_URL="https://demo-cicd.cloud.gooddata.com"
```
hkad98, line 11: Do we want to keep this?

jaceksan: I vote for keeping it there until we introduce a better solution, as an example.
I tried to make it configurable from the outside, but so far I failed.

hkad98 (approved):

PR #726, opened 2024-07-16
RELATED: CQ-555 - Add new gooddata-flight-server package

lupko:
This new package delivers:

- templated / pluggable Flight RPC server skeleton
- skeleton takes care of:
  - configuration according to 12-factor app principles
  - maintenance
  - health checks
  - logging
  - metrics
  - tracing
  - long-running tasks
  - TLS/mTLS setup
  - startup procedure & graceful shutdown

From the outside, it is possible to plug in the implementation of Flight RPC methods - the rest is taken care of by the template.

Additionally, the server comes with default implementation of Flight RPC methods: the FlexFun. This takes things even further and establishes an interface for custom functions that generate data; these functions are exposed via Flight RPC.


gooddata-flight-server/Makefile
```diff
@@ -0,0 +1,6 @@
+# (C) 2021 GoodData Corporation
```
hkad98, line 1: Nitpick: check copyright dates pls

gooddata-flight-server/setup.py
```diff
@@ -0,0 +1,69 @@
+# (C) 2021 GoodData Corporation
+from pathlib import Path
+
+from setuptools import find_packages, setup
+
+this_directory = Path(__file__).parent
+long_description = (this_directory / "README.md").read_text(encoding="utf-8")
+
+REQUIRES = [
```
no23reason, line 9: Can this be taken from the requirements.txt somehow? What would happen if this and requirements.txt fell out of sync?

gooddata-flight-server/gooddata_flight_server/config/config.py
```diff
@@ -0,0 +1,515 @@
...
+    OtelExporterType.Console.value,
+]
+
+_SUPPORTED_AUTH_METHOD = [
+    AuthenticationMethod.NoAuth.value,
+    AuthenticationMethod.Token.value,
+]
+
+
+def _s(name: str) -> str:
```
hkad98, line 153: Is this function name correct?

gooddata-flight-server/gooddata_flight_server/config/config.py
```diff
@@ -0,0 +1,515 @@
...
+        with open(value[1:], "rb") as f:
+            return f.read()
+
+    return value.encode("ascii")
+
+
+def _create_server_config(settings: Dynaconf) -> ServerConfig:
+    server_settings = settings.get(_SERVER_SECTION_NAME)
+
+    exporter_type = server_settings.get(_Settings.OtelExporterType)
```
hkad98, line 423: server_settings can be None, in that case `server_settings.get(...)` fails

gooddata-flight-server/gooddata_flight_server/server/auth/auth_middleware.py
```diff
@@ -0,0 +1,82 @@
...
+            return values[0]
+
+        if self._token_header_name is None:
+            token = _auth_header_value(_DEFAULT_AUTH_TOKEN_HEADER)
+            if not token.startswith("Bearer "):
+                raise pyarrow.flight.FlightUnauthenticatedError(
+                    "Authentication failed because the 'Authorization' header does not start with 'Bearer '"
+                )
+
+            return token[7:].strip()
```
no23reason, line 66: Can we specify the 7 as something related to the "Bearer" prefix (I suppose that is it?)

gooddata-flight-server/setup.py
lupko: that's how it's done in all packages... this is the 'old style setup'.

gooddata-flight-server/gooddata_flight_server/config/config.py
lupko: fixed

gooddata-flight-server/gooddata_flight_server/cli.py
```diff
@@ -0,0 +1,132 @@
...
+        "--config",
+        type=str,
+        nargs="*",
+        metavar="CONFIG_FILE",
+        help="Optionally specify one or more setting files to use. If not specified, all configuration has to be "
+        "provided using environment variables. Important: if you only specify this optional argument, "
+        "you must use the `--config-end` to explicitly terminate the list of files - otherwise the CLI cannot "
+        "correctly distinguish the end of list and will fail with usage error.",
+    )
+    start_cmd.add_argument(
```
no23reason, line 38: This is a pattern I haven seen before :-/ why not accept multiple --config values?

gooddata-flight-server/gooddata_flight_server/config/config.py
lupko: yep, it's intentionally short.. it is used internally & repetetivly when constructing dynaconf validators. it wraps around the setting name, adds the section.

gooddata-flight-server/Makefile
lupko: went through all copyrights and bumped them to current year

gooddata-flight-server/gooddata_flight_server/cli.py
lupko:
```
    start_cmd.add_argument(
        "--config",
        type=str,
        nargs="*",
        metavar="CONFIG_FILE",
```

the --config is already setup with '*' number of arguments. the --config-end exists because otherwise the argparse may have problems going through something like this:

./cli --config a.toml b.toml <some other positional argument>

we do not use the positional arg now in this CLI, so this arg may be removed (this was from Quiver where we had say `./cli --config ... start --node something`)

gooddata-flight-server/gooddata_flight_server/server/auth/auth_middleware.py
lupko: added constant

gooddata-flight-server/gooddata_flight_server/config/config.py
lupko: i have renamed it to `_fqsn` -> fully qualified setting name + added doc

PR #746, opened 2024-07-25
refactor: adapt to the change of NATIVE -> NAT on backend

no23reason:
To keep things consistent, we changed the TotalFunction enum value for native totals from NATIVE to NAT on the backend. This is the adaptation to that.

JIRA: CQ-652
risk: low


lupko (approved):
LGTM

PR #762, opened 2024-08-11
Drop Python 3.8 support

hkad98:


CONTRIBUTING.md
```diff
@@ -2,7 +2,7 @@
 
 ## Getting Started
 
-1. Ensure you have at minimum Python 3.12 installed; Python 3.10, 3.9 and 3.8 are optional for multi-environment tests
+1. Ensure you have at minimum Python 3.12 installed; Python 3.10 and 3.9 are optional for multi-environment tests
```
no23reason, line 5: Missing 3.11

gooddata-fdw/gooddata_fdw/import_workspace.py
```diff
@@ -63,15 +63,15 @@ def import_tables(self) -> list[TableDefinition]:
 
 
 class WorkspaceImportersLocator:
-    _IMPORTERS: set[Type[WorkspaceImporter]] = set()
+    _IMPORTERS: set[type[WorkspaceImporter]] = set()
 
     @classmethod
-    def register(cls, class_: Type[WorkspaceImporter]) -> Type[WorkspaceImporter]:
+    def register(cls, class_: type[WorkspaceImporter]) -> type[WorkspaceImporter]:
```
no23reason, line 69: This typing is imprecise. This means you pass it _some_ `type[WorkspaceImporter]` and it returns (possibly some other) `type[WorkspaceImporter]`.

The register method should be generic on this with a TypeVar bound to `type[WorkspaceImporter]`.

no23reason: Not super critical though, it does not make the status quo worse ;)

pyproject.toml
```diff
@@ -21,11 +21,14 @@ lint.select = [
...
-    "W" # pycodestyle warnings
+    "UP006",
+    "W", # pycodestyle warnings
 ]
 lint.ignore = [
     # rules recommended to be ignored when using ruff format
-    "E501", "W191", "W291", "W293"
+    "E501", "W191",
+    # ignore the rule banning Optional[X] in favor of X | None, we are fine with Optional for now
+    "UP007",
```
no23reason, line 31: No need to ignore this, we do no not enable the whole UP. Or maybe we should enable the whole UP and not only UP006?

gooddata-sdk/gooddata_sdk/catalog/workspace/model_container.py
```diff
@@ -126,7 +126,7 @@ def _valid_objects(self, ctx: ValidObjectsInputType) -> ValidObjects:
 
     def filter_by_valid_objects(
         self, valid_objects: dict[str, set[str]]
-    ) -> tuple[List[CatalogDataset], List[CatalogMetric]]:
+    ) -> tuple[list[CatalogDataset], list[CatalogMetric]]:
```
no23reason, line 129: Hm, there might be some other uses of `List` in this file, I don't see it removed from the imports.

gooddata-fdw/gooddata_fdw/import_workspace.py
hkad98: Agree. gooddata_fdw is currently marked as deprecated. I propose to keep it as is.

pyproject.toml
hkad98: I want to play with this in the follow-up. I commented UP007 because it is commented on in Quiver. I plan to check other UP rules, but first I need to check if it is required to have python-version set up to the minimum supported one.

CONTRIBUTING.md
hkad98: Missing, added.

gooddata-sdk/gooddata_sdk/catalog/workspace/model_container.py
hkad98: Indeed. Removed.

no23reason (dismissed):

no23reason (dismissed):

no23reason (approved):

PR #763, opened 2024-08-12
feat: add convertor from Compute dicts to SDK model

no23reason:
This will be useful in the Flight RPC functions as they will get the execution as a JSON from the server and we want to make them more ergonomic by converting them to the SDK model (e.g. to launch another execution from the SDK).

JIRA: CQ-577
risk: low


hkad98:
Could you please add docstring to `compute_to_sdk_converter.py`? It might be misleading why it is so similar to _convert_filter_to_computable, where, for example, RelativeDateFilter has `dataSet`, but sdk_convereter has `dataset`.

hkad98 (approved):

PR #765, opened 2024-08-13
feat: add classes to parse executionRequest in Flight RPC

no23reason:
Allow the users to parse the incoming executionRequest when invoking the Flight RPC function. Use the gooddata-sdk objects so that they can be easily reused in the gooddata-sdk itself.

Also, add the gooddata-flight-server to the CI pipelines, make sure the coverage is published and fix all warnings in the Dockerfile.

Fix some code smells along the way, see the last commit for more details.

JIRA: CQ-577
risk: low

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,8 +27,8 @@ RUN set -x \
   && true
 
 # install tox
-ENV PYTHON_TOX_VERSION=4.14.0
-ENV PYTHON_TOX_UV_VERSION=1.5.1
+ENV PYTHON_TOX_VERSION=4.14.1
+ENV PYTHON_TOX_UV_VERSION=1.7.0
 RUN set -x \
   && pip3 install tox==${PYTHON_TOX_VERSION} tox-uv==${PYTHON_TOX_UV_VERSION}\
   && true
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,13 +2,14 @@
 ARG PY_TAG
 FROM python:${PY_TAG}
 
+ARG PY_TAG
 ARG ENV_TAG
 # tox defines all python targets, makefile recognizes TEST_ENVS and forces
 # tox to execute only tests for installed python
 ENV TEST_ENVS=${ENV_TAG}
 
 # install make and gosu
-ENV GOSU_VERSION 1.14
+ENV GOSU_VERSION=1.14
 RUN set -x \
   && apt-get update \
   && apt-get install -y --no-install-recommends make curl gnupg \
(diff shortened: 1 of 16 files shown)


Dockerfile
```diff
@@ -26,8 +27,8 @@ RUN set -x \
   && true
 
 # install tox
-ENV PYTHON_TOX_VERSION 4.14.0
-ENV PYTHON_TOX_UV_VERSION 1.5.1
+ENV PYTHON_TOX_VERSION=4.14.0
+ENV PYTHON_TOX_UV_VERSION=1.5.1
```
hkad98, line 31: Pls, bump tox to 4.14.1 and tox uv to 1.7.0. So it is in sync with tox-requirements.txt

gooddata-flight-server/gooddata_flight_server/flexfun/flex_fun_execution_context.py
```diff
@@ -0,0 +1,105 @@
+# (C) 2024 GoodData Corporation
+from dataclasses import dataclass
+from typing import Optional
+
+from gooddata_sdk import Attribute, ComputeToSdkConverter, Filter, Metric
+
+
+@dataclass
```
hkad98, line 8: Consider using attrs instead.

gooddata-flight-server/gooddata_flight_server/flexfun/flex_fun_execution_context.py
```diff
@@ -0,0 +1,105 @@
...
+        :param d: the dictionary to parse
+        """
+        return ExecutionRequest(
+            attributes=[ComputeToSdkConverter.convert_attribute(a) for a in d.get("attributes", [])],
+            metrics=[ComputeToSdkConverter.convert_metric(m) for m in d.get("measures", [])],
+            filters=[ComputeToSdkConverter.convert_filter(f) for f in d.get("filters", [])],
+        )
+
+
+@dataclass
```
hkad98, line 42: Consider using attrs instead.

Dockerfile
no23reason: updated

gooddata-flight-server/gooddata_flight_server/flexfun/flex_fun_execution_context.py
no23reason: Let's not add this dependency until we need some of its features

gooddata-flight-server/gooddata_flight_server/flexfun/flex_fun_execution_context.py
no23reason: Let's not add this dependency until we need some of its features

hkad98 (approved):

PR #770, opened 2024-08-15
Enable some rules for ruff

hkad98:


gooddata-flight-server/gooddata_flight_server/tasks/task.py
```diff
@@ -104,10 +105,8 @@ def cancel(self) -> bool:
             self._cancelled = True
 
             if first_cancel:
-                try:
+                with contextlib.suppress(Exception):
```
no23reason, line 108: This will make the happy path slower every time.

hkad98: I added the relevant rule SIM105 to ignore.

no23reason (approved):

PR #788, opened 2024-09-10
refactor: split gooddata-flight-server

no23reason:
Split GoodData FlexFun-specific stuff into a separate package. This leaves gooddata-flight-server to be a GoodData-agnostic server implementation better encapsulating the responsibilities.

Also make tox.ini files consistent with regards to the supported python versions and make sure the development testing certificates last longer.

JIRA: CQ-754
risk: low


no23reason:
Closing, the rebasing would be too much effort, will do the split again on current master branch.

PR #793, opened 2024-09-11
feat: add support for more label elements parameters

no23reason:
Add support for paging, pattern filters, sort order and complementary filters to the collect label elements.

One of the test cases had to be removed as the up-to-date test model does not have any suitable attribute anymore.

JIRA: CQ-764
risk: low

### DIFF
--- a/gooddata-sdk/gooddata_sdk/catalog/workspace/content_service.py
+++ b/gooddata-sdk/gooddata_sdk/catalog/workspace/content_service.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import copy
 import functools
 from pathlib import Path
-from typing import Optional, Union
+from typing import Literal, Optional, Union
 
 import gooddata_api_client.models as afm_models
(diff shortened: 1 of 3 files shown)


hkad98 (approved):

gooddata-sdk/gooddata_sdk/catalog/workspace/content_service.py
```diff
@@ -612,6 +628,23 @@ def get_label_elements(
...
+        if sort_order is not None:
+            request.sort_order = sort_order
+
+        paging_params = {}
+        if offset is not None:
+            paging_params["offset"] = offset
+        if limit is not None:
+            paging_params["limit"] = limit
+
         # TODO - fix return type of Paging.next in Backend + add support for this API to SDK
```
hkad98, line 646: Is this comment still valid?

no23reason: Unfortunately it is. Not sure how to fix that at all...

PR #798, opened 2024-09-12
feat: extend FlexFun execution context

no23reason:
- differentiate between report and label elements executions
- deprecate the generic ExecutionRequest
- get rid of the `attrs` package, it is not worth the hassle
- add some tests

JIRA: CQ-764
risk: low


lupko (approved):
LGTM

PR #803, opened 2024-09-13
feat: split the gooddata-flight-server into two pacakges

no23reason:
This makes the gooddata-flight-server into a generic FlightRPC server with no dependencies to anything GoodData-specific. The GoodData FlexFun parts were separated into a separate package.
To re-connect those two a mechanism for loading the server methods was added.

JIRA: CQ-754, CQ-756
risk: low


gooddata-flight-server/gooddata_flight_server/server/base.py
```diff
@@ -61,3 +62,19 @@ class FlightServerMethodsFactory(Protocol):
     """
 
     def __call__(self, ctx: ServerContext) -> FlightServerMethods: ...
+
+
+class FlightServerMethodsAbstractFactory(abc.ABC):
```
lupko, line 67: the naming here is not correct because this is not actually an abstract factory. eg: https://en.wikipedia.org/wiki/Abstract_factory_pattern

i suggest to go slightly differently in this case; instead of doing 'factory for factory', the server could expose a decorator (used really as a marker) that would be slapped on the existing factory function. the dynamic loading code would search for that decorator.

i don't have experience with this approach but it is definitely possible - other libraries do it too.

simplest alternative is a convention where the factory function must have particular name. the server will look for that function and will fail upon startup otherwise.

no23reason: How is that not an abstract factory? It is a class that returns factories 🤔 
Nevertheless, I like the marker decorator pattern, will give it a go.

no23reason: Rewrote to using decorators.

gooddata-flight-server/gooddata_flight_server/utils/methods_discovery.py
```diff
@@ -0,0 +1,58 @@
...
+
+    methods: list[FlightServerMethodsFactory] = [
+        member.__wrapped__  # unwrap the actual instance from the marker decorator to keep the type
+        for member in module.__dict__.values()
+        if _is_flight_server_methods(member)
+    ]
+
+    if len(methods) != 1:
+        raise ValueError(
+            f"Expected exactly one FlightServerMethodsFactory decorated with @flight_server_methods in {module_name}, got {len(methods)}"
```
lupko, line 55: imho this error message should be expanded and be more 'beginner' friendly. also, the flight-server code should ensure that when a particular type of error happens when calling the discovered factory function, then the message is printed out 'nicely'; without stack-trace and other nasty stuff.

it may be better to have a dedicated exception (flight server may already have some that can be reused?) for these types of errors - to avoid scenario when ValueError flies due to genuine bug/error and our code then tosses away the stacktrace.

+ i suggest also cover (via error handling/messaging) the case when someone codes - by mistake and against the contract - factory function with less or more parameters (thus leading to errors).

+ perhaps the scenario when the decorated function returns nothing (again, error handling / messaging only).

---

our messaging in these cases needs to be.. you made a mistake coding the function. you should only annotate a function that has the right signature. correct your code, check out docs, blabla.

PR #805, opened 2024-09-17
chore: remove flight-server dependencies

no23reason:
Flight server no longer depends on SDK and API client, so let's remove that from tox.

Also add missing dev-requirements.txt entries for Flexfun.

JIRA: CQ-754
risk: low


lupko (approved):
LGTM

PR #834, opened 2024-10-07
Improve GoodPandas performance 

hkad98:
Noticed that it is not necessary to call heavy `get_full_catalog`. `get_attributes_catalog` can be called instead.

JIRA: PSDK-205
risk: low


gooddata-pandas/gooddata_pandas/data_access.py
```diff
@@ -16,6 +16,7 @@
     ObjId,
     TableDimension,
 )
+from gooddata_sdk.utils import IdObjType
```
no23reason, line 19: Is this import correct?

hkad98: Yes, before that [catalog.find_label_attribute](https://github.com/hkad98/gooddata-python-sdk/blob/3bee712c76dced06be8e46264a12218306cf1af9/gooddata-sdk/gooddata_sdk/catalog/workspace/entity_model/content_objects/dataset.py#L166-L166) was used and based on that typing the same import was used [see](https://github.com/hkad98/gooddata-python-sdk/blob/3bee712c76dced06be8e46264a12218306cf1af9/gooddata-sdk/gooddata_sdk/catalog/workspace/entity_model/content_objects/dataset.py#L17-L17).

no23reason (approved):

PR #856, opened 2024-10-21
fix: export_tabular_by_visualization_id does not need to call afm

hkad98:
It is not necessary to execute visualization and get `execution_result` for `export_tabular_by_visualization_id`. Now export by visualization_id is supported out of the box.

JIRA: TRIVIAL
risk: low

### DIFF
--- a/gooddata-sdk/gooddata_sdk/catalog/export/service.py
+++ b/gooddata-sdk/gooddata_sdk/catalog/export/service.py
@@ -9,15 +9,10 @@ from gooddata_api_client.model.visual_export_request import VisualExportRequest
 
 from gooddata_sdk.catalog.catalog_service_base import CatalogServiceBase
 from gooddata_sdk.catalog.export.request import (
-    ExportCustomLabel,
-    ExportCustomMetric,
-    ExportCustomOverride,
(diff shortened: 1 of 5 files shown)


gooddata-sdk/gooddata_sdk/catalog/export/service.py
```diff
@@ -321,15 +321,12 @@ def export_tabular_by_visualization_id(
...
-        custom_override = self._custom_overrides_labels(exec_table)
-        file_name = file_name if file_name is not None else visualization_tile
+        file_name = file_name if file_name is not None else "export"
         export_request = ExportRequest(
             format=file_format,
-            execution_result=exec_table.result_id,
+            visualization_object=visualization_id,
             file_name=file_name,
             settings=settings,
-            custom_override=custom_override,
```
Vojtasii, line 332: Is this not important?

Vojtasii: It's now done in tabular-exporter itself, so we are fine

Vojtasii (dismissed):

Vojtasii (approved):

PR #858, opened 2024-10-23
fix: regenerate API client and fix automations

hkad98:
There was an issue with DeclarativeNotificationChannel. It was fixed on backend and it needed to be propagated to PySDK. Also, there were some issues with Automation entity that were fixed.

JIRA: TRIVIAL
risk: low


gooddata-sdk/gooddata_sdk/catalog/export/request.py
```diff
@@ -96,3 +97,27 @@ def file(self) -> str:
             str: Full filename with the format extension.
         """
         return f"{self.file_name}.{self.format.lower()}"
+
+
+@define(auto_attribs=True, kw_only=True)
+class VisualExportRequest(Base):
```
Vojtasii, line 103: VisualExportRequest addded, but not TabularExportRequest? Is it elsewhere?

Vojtasii: I see now, ExportRequest is TabularExportRequest

Vojtasii (approved):

PR #878, opened 2024-11-12
feat: FilterView management in pythonSDK

chrisbonilla95:
Enable the management of FilterViews from the pythonSDK

JIRA: LX-428
risk: low

### DIFF
--- a/gooddata-sdk/gooddata_sdk/catalog/identifier.py
+++ b/gooddata-sdk/gooddata_sdk/catalog/identifier.py
@@ -7,6 +7,9 @@ import attr
 from attrs import define
 from gooddata_api_client.model.assignee_identifier import AssigneeIdentifier
 from gooddata_api_client.model.dataset_workspace_data_filter_identifier import DatasetWorkspaceDataFilterIdentifier
+from gooddata_api_client.model.declarative_analytical_dashboard_identifier import (
+    DeclarativeAnalyticalDashboardIdentifier,
+)
 from gooddata_api_client.model.declarative_export_definition_identifier import DeclarativeExportDefinitionIdentifier
 from gooddata_api_client.model.declarative_notification_channel_identifier import (
     DeclarativeNotificationChannelIdentifier,
@@ -114,3 +117,13 @@ class CatalogNotificationChannelIdentifier(Base):
     @staticmethod
     def client_class() -> builtins.type[DeclarativeNotificationChannelIdentifier]:
         return DeclarativeNotificationChannelIdentifier
+
+
+@attr.s(auto_attribs=True, kw_only=True)
+class CatalogDeclarativeAnalyticalDashboardIdentifier(Base):
+    id: str
+    type: str = attr.field(validator=value_in_allowed)
+
+    @staticmethod
+    def client_class() -> builtins.type[DeclarativeAnalyticalDashboardIdentifier]:
+        return DeclarativeAnalyticalDashboardIdentifier
--- a/gooddata-sdk/gooddata_sdk/catalog/workspace/entity_model/filter_view.py
+++ b/gooddata-sdk/gooddata_sdk/catalog/workspace/entity_model/filter_view.py
@@ -1,0 +1,146 @@
+# (C) 2024 GoodData Corporation
+from __future__ import annotations
+
+from typing import Any, Optional, Union
+
+import attr
+from gooddata_api_client.model.json_api_filter_view_in import JsonApiFilterViewIn
+from gooddata_api_client.model.json_api_filter_view_in_attributes import JsonApiFilterViewInAttributes
+from gooddata_api_client.model.json_api_filter_view_in_document import JsonApiFilterViewInDocument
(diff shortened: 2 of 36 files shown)


gooddata-sdk/gooddata_sdk/client.py
```diff
@@ -57,6 +57,7 @@ def __init__(
         self._layout_api = apis.LayoutApi(self._api_client)
         self._actions_api = apis.ActionsApi(self._api_client)
         self._user_management_api = apis.UserManagementApi(self._api_client)
+        self._filter_views_api = apis.FilterViewsApi(self._filter_views_api)
```
hkad98, line 60: I believe that this is not necessary. `FilterViewsApi` should be part of `self._layout_api ` or `self._entities_api` or are they not? There is `self._user_management_api` because these are use case api.

gooddata-sdk/gooddata_sdk/catalog/identifier.py
```diff
@@ -82,7 +85,7 @@ def client_class() -> builtins.type[DeclarativeUserIdentifier]:
 @attr.s(auto_attribs=True, kw_only=True)
 class CatalogLabelIdentifier(Base):
     id: str
-    type: str = attr.field(validator=value_in_allowed)
+    type: str = attr.field(default="analyticalDashboard", validator=value_in_allowed)
```
hkad98, line 88: Is the default correct here? If I check [LabelIdentifier](https://github.com/gooddata/gooddata-python-sdk/blob/3f112f4641aaa27b13d1f60a4ef5ab9e57539897/gooddata-api-client/gooddata_api_client/model/label_identifier.py#L58-L58) client class there is only `label` in allowed_values.

chrisbonilla95: This looks like it was introduced when merging from latest and resolving conflicts. I did not modify this, let me revert.

hkad98:
I do not see any tests. Could you please add some?

gooddata-sdk/gooddata_sdk/catalog/workspace/entity_model/filter_view.py
```diff
@@ -0,0 +1,147 @@
...
+        return JsonApiFilterViewIn(id=self.id, attributes=attributes, relationships=relationships)
+
+    @property
+    def user_id(self) -> str | None:
+        if self.relationships and self.relationships.user:
+            return self.relationships.user["data"].id
+        return None
+
+    @property
+    def analytical_dashboard_id(self) -> str | None:
```
hkad98, line 79: `str | None` is supported since 3.10. See [PEP-604](https://peps.python.org/pep-0604/). Strangely, mypy did not catch this. Please, use `Union[str, None] `instead.

gooddata-sdk/gooddata_sdk/catalog/workspace/entity_model/filter_view.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def assign_analytical_dashboard(self, analytical_dashboard_id: str) -> None:
+        if self.relationships is None:
+            self.relationships = CatalogFilterViewRelationships.create_user_analytical_dashboard_relationship(
+                analytical_dashboard_id=analytical_dashboard_id
+            )
+        else:
+            self.relationships.analytical_dashboard = _data_entity(CatalogEntityIdentifier(id=analytical_dashboard_id))
+
+    def clean_assignments(self) -> None:
```
hkad98, line 100:
```suggestion
    def clean_relationships(self) -> None:
```

gooddata-sdk/gooddata_sdk/catalog/workspace/entity_model/filter_view.py
```diff
@@ -0,0 +1,147 @@
...
+            if analytical_dashboard_id
+            else None
+        )
+        return cls(user=assignee_user, analytical_dashboard=assignee_analytical_dashboard)
+
+
+@attr.s(auto_attribs=True, kw_only=True)
+class CatalogEntityIdentifier(Base):
+    id: str
+    type: Optional[str] = None
```
hkad98, line 147: I am not sure if CatalogEntityIdentifier is a good idea. There are identifiers in identfier.py Could we reuse them?

gooddata-sdk/gooddata_sdk/catalog/identifier.py
```diff
@@ -114,3 +117,13 @@ class CatalogNotificationChannelIdentifier(Base):
     @staticmethod
     def client_class() -> builtins.type[DeclarativeNotificationChannelIdentifier]:
         return DeclarativeNotificationChannelIdentifier
+
+
+@attr.s(auto_attribs=True, kw_only=True)
+class CatalogDeclarativeAnalyticalDashboardIdentifier(Base):
+    id: str
+    type: str = attr.field(validator=value_in_allowed)
```
hkad98, line 125: If there is only one type in `DeclarativeAnalyticalDashboardIdentifier` then it is not necessary to have type here. See identifers above.

hkad98 (changes requested):

hkad98 (approved):

PR #952, opened 2025-01-17
perf: rewrite data access in gooddata-pandas

hkad98:
data_access.py had poor performance because it loaded all attributes. The new implementation fetches only the necessary ones which should perform better on bigger workspaces.

JIRA: STL-1036
risk: low


gooddata-sdk/gooddata_sdk/utils.py
```diff
@@ -424,3 +425,8 @@ def ref_extract(ref: dict[str, Any]) -> Union[str, ObjId]:
         return ref["localIdentifier"]
 
     raise ValueError("invalid ref. must be identifier or localIdentifier")
+
+
+def filter_for_attributes_labels(attributes: list[Attribute]) -> str:
+    label_ids = [attribute.label.id for attribute in attributes]
```
no23reason, line 431: Would make this a set, there can be duplicates in the execution, but no need to include them multiple times.

gooddata-pandas/gooddata_pandas/data_access.py
```diff
@@ -446,7 +446,10 @@ def compute_and_extract(
     if not exec_def.has_attributes():
         return _extract_for_metrics_only(response, cols, col_to_metric_idx), dict()
     else:
-        attributes = sdk.catalog_workspace_content.get_attributes_catalog(workspace_id, include=["labels", "datasets"])
+        filter_query = filter_for_attributes_labels(exec_def.attributes)
+        attributes = sdk.catalog_workspace_content.get_attributes_catalog(
+            workspace_id, include=["labels", "datasets"], rsql_filter=filter_query
```
no23reason, line 451: Are we sure this will all fit into URL not hitting the length limit?

hkad98: It is a good point. I tried it, and it really can hit the limitation.

```
gooddata_api_client.exceptions.ApiException: Status Code: 414
Reason: Request-URI Too Large
HTTP response headers: HTTPHeaderDict({'Server': 'nginx', 'Date': 'Fri, 17 Jan 2025 13:07:20 GMT', 'Content-Type': 'text/html; charset=UTF-8', 'Connection': 'close'})
HTTP response body: <html>
<head><title>414 Request-URI Too Large</title></head>
<body>
<center><h1>414 Request-URI Too Large</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

hkad98: The limit for URLs is 2048 characters. I will limit rsql filter to 1500 characters. If it is higher, it will fetch all attributes.

gooddata-sdk/gooddata_sdk/utils.py
hkad98: Good point. I tried to solve it as well. Unfortunately, `set()` cannot be used because the order is not deterministic, but I found a way to remove duplicates.

no23reason (approved):

PR #966, opened 2025-02-05
feat: make api docs visible from docker compose

Mara3l:


docs/Dockerfile
```diff
@@ -1,14 +1,35 @@
...
     apt-get install -y git make golang-go curl && \
     npm install -g hugo-extended@0.117.0 && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-COPY docs docs
-
-WORKDIR docs
+WORKDIR /docs
 RUN npm install
```
hkad98, line 33: Please, add `hugo mod get` so modules are downloaded during build.

docs/Dockerfile
```diff
@@ -1,14 +1,35 @@
+FROM python:3.11-slim AS builder
```
hkad98, line 1: Use Python 3.12, please.

docs/Dockerfile
```diff
@@ -1,14 +1,35 @@
...
+RUN apt-get update && apt-get install -y git make curl
+
+COPY scripts/script-requirements.txt /scripts/script-requirements.txt
+COPY docs docs
+COPY scripts/docs/ /docs
+COPY gooddata-api-client /gooddata-api-client
+COPY gooddata-sdk /gooddata-sdk
+COPY gooddata-pandas /gooddata-pandas
+
+RUN pip install -r /scripts/script-requirements.txt
```
hkad98, line 12: `RUN pip install --no-cache-dir -r /scripts/script-requirements.txt`

PR #1020, opened 2025-04-15
feat: cancel tokens bouncing

fiedlr:
risk: high

### DIFF
--- a/gooddata-sdk/gooddata_sdk/compute/service.py
+++ b/gooddata-sdk/gooddata_sdk/compute/service.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import logging
 
+from gooddata_api_client.model.afm_cancel_tokens import AfmCancelTokens
 from gooddata_api_client.model.chat_history_request import ChatHistoryRequest
 from gooddata_api_client.model.chat_history_result import ChatHistoryResult
 from gooddata_api_client.model.chat_request import ChatRequest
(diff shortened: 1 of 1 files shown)


gooddata-sdk/gooddata_sdk/compute/service.py
```diff
@@ -34,13 +34,16 @@ def for_exec_def(self, workspace_id: str, exec_def: ExecutionDefinition) -> Exec
...
+        response, _, http_headers = self._actions_api.compute_report(
+            workspace_id, exec_def.as_api_model(), _check_return_type=False
+        )
 
         return Execution(
             api_client=self._api_client,
             workspace_id=workspace_id,
             exec_def=exec_def,
             response=response,
+            cancel_token=http_headers["X-GDC-CANCEL-TOKEN"] if exec_def.is_cancellable else None,
```
hkad98, line 46: This fails when `X-GDC-CANCEL-TOKEN` is not included in http_headers and exec_def.is_cancellable is True. Consider using `http_headers.get("X-GDC-CANCEL-TOKEN")` if key is not presented then None is used.

gooddata-sdk/gooddata_sdk/compute/model/execution.py
```diff
@@ -335,6 +347,7 @@ def read_result(self, limit: Union[int, list[int]], offset: Union[None, int, lis
             limit=_limit,
             _check_return_type=False,
             _return_http_data_only=False,
+            x_gdc_cancel_token=self.cancel_token,
```
hkad98, line 350: Be careful here. x_gdc_cancel_token should be string, but there is a difference between Python string boolean ("True", "False") and other string boolean representation ("true", "false"). Make sure that it will work.

no23reason (approved):

hkad98 (approved):

PR #1065, opened 2025-06-12
feat: add on_execution_submitted to gooddata_pandas

no23reason:
Users can now provide a callback that will be called with the Execution object as soon as it is available. This is useful for getting some of the information available in the Execution object before the actual data is loaded. The Execution object can also be used to cancel the execution result operation.

Also unify Execution types in data_access.py: no need to use the ExecutionResponse alias anymore.

JIRA: CQ-1387
risk: low

### DIFF
--- a/gooddata-pandas/gooddata_pandas/data_access.py
+++ b/gooddata-pandas/gooddata_pandas/data_access.py
@@ -1,12 +1,13 @@
 # (C) 2021 GoodData Corporation
 from __future__ import annotations
 
-from typing import Any, Optional, Union
+from typing import Any, Callable, Optional, Union
 
 from gooddata_sdk import (
     Attribute,
     AttributeFilter,
     CatalogAttribute,
+    Execution,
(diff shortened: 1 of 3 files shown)


gooddata-pandas/gooddata_pandas/data_access.py
```diff
@@ -441,6 +444,9 @@ def compute_and_extract(
 
     response, col_to_attr_idx, col_to_metric_idx, index_to_attr_idx = result
 
+    if on_execution_submitted is not None:
+        on_execution_submitted(response)
```
lupko, line 448: the `response` name is now not well-matching. would be better to do mild refactors?

lupko (approved):
LGTM

gooddata-pandas/gooddata_pandas/data_access.py
no23reason: good point, changed!

PR #1073, opened 2025-06-18
feat: add is_cancellable support to GoodPandas

no23reason:
As already done in GoodData SDK, GoodPandas now also supports this parameter with the same default.

JIRA: CQ-1411
risk: low

### DIFF
--- a/gooddata-pandas/gooddata_pandas/data_access.py
+++ b/gooddata-pandas/gooddata_pandas/data_access.py
@@ -33,7 +33,7 @@ from gooddata_pandas.utils import (
 class ExecutionDefinitionBuilder:
     _DEFAULT_INDEX_NAME: str = "0"
 
-    def __init__(self, columns: ColumnsDef, index_by: Optional[IndexDef] = None) -> None:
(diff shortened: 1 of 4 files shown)


gooddata-pandas/gooddata_pandas/data_access.py
```diff
@@ -42,13 +42,16 @@ def __init__(self, columns: ColumnsDef, index_by: Optional[IndexDef] = None) ->
         Args:
             columns (ColumnsDef): Input columns to process and build internal mappings.
             index_by (Optional[IndexDef], optional): Index definition to process. Defaults to None.
+            is_cancellable (Optional[bool]): Whether the execution of this definition should be cancelled when
+                the connection is cancelled.
```
lupko, line 46: small stuff with wording; the "connection is interrupted" instead of "connection is cancelled". connection cancellation is not a commonly used term i believe

no23reason: good point, reworded

lupko (approved):
LGTM

PR #1110, opened 2025-08-13
feat(gooddata-pipelines): Backup workspaces from list of workspace IDs

benkeanna:


gooddata-pipelines/gooddata_pipelines/backup_and_restore/backup_input_processor.py
```diff
@@ -158,25 +158,48 @@ def get_all_workspaces(self) -> list[str]:
...
-            if path_to_csv is None:
+            if path_to_csv is None and workspace_ids is None:
                 raise ValueError(
-                    f"Path to CSV is required for this input type: {input_type.value}"
+                    f"Path to CSV or list of workspace IDs is required for this input type: {input_type.value}"
+                )
+            elif path_to_csv is not None and workspace_ids is not None:
+                raise ValueError(
+                    f"Path to CSV and list of workspace IDs are mutually exclusive for this input type: {input_type.value}"
                 )
```
hkad98, line 176: The condition can be simplified.

```
# Both or none of them is specified.
if (path_to_csv is None) == (workspace_ids is None):
```

PR #1125, opened 2025-09-02
feat: optimize memory allocation when converting execution response to dataframe

Martozar:
Add `optimized` flag to DataFrameFactory to enable memory-optimized conversion of execution response to pandas dataframe. Without the flag, the conversion will run as usual, storing headers as a list of dictionaries. The optimized version only stores unique headers and references them, thereby preventing unnecessary memory allocations when processing large numbers of duplicated headers. On large datasets, optimized conversation consumes up to 10x less memory (e.g. on 1M rows with 4 attributes original implementation consumed almost 2Gb, while optimized not more than 200Mb)

Note that the new behaviour is optional and turned off by default, so no existing usages should be affected.

JIRA: CQ-1579
risk: low

### DIFF
--- a/gooddata-pandas/gooddata_pandas/result_convertor.py
+++ b/gooddata-pandas/gooddata_pandas/result_convertor.py
@@ -1,4 +1,7 @@
 # (C) 2022 GoodData Corporation
+from abc import ABC, abstractmethod
+from collections.abc import Iterator
+from functools import cached_property
 from typing import Any, Callable, Optional, Union, cast
 
 import pandas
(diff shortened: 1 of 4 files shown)


gooddata-pandas/gooddata_pandas/result_convertor.py
```diff
@@ -11,14 +14,179 @@
 LabelOverrides = dict[str, dict[str, dict[str, str]]]
 
 
+class _Header(ABC):
```
lupko, line 17: i think this should also have `__slots__ = ()` so as not to pollute class hierarchy.

see: https://stackoverflow.com/questions/1816483/how-does-inheritance-of-slots-in-subclasses-actually-work

gooddata-pandas/gooddata_pandas/result_convertor.py
```diff
@@ -11,14 +14,174 @@
...
+class _HeaderContainer:
+    """
+    Container for headers to improve memory usage.
+    Unique headers are stored as keys in _header_to_header and references to them are stored in _headers.
+    This way we avoid storing the same header multiple times, reducing memory allocations,
+    which is important for large datatables with many attributes.
+    """
+
+    _headers: list[_Header] = field(factory=list)
+    _header_to_header: dict[_Header, _Header] = field(factory=dict)
```
lupko, line 129: i'd rename this. `_header_cache` or `_header_dedup` or something like that. the `header to header` sounds like a name of some pop band :)

Martozar: Thanks, fixed :)

lupko (approved):
LGTM

PR #1164, opened 2025-09-25
feat: upgrade OpenAPI Generator 6.0.1 -> 7.11.0

no23reason:
This is not the latest version, but a version that does not  significantly change the generated types (newer versions lead to some significant renames which make the change even bigger).

Changes needed:
- change `from gooddata_api_client.model.` to `from gooddata_api_client.models.`
- change `from gooddata_api_client import apis` to `from gooddata_api_client import api`
- remove `_preload_content`, there are now special methods with `_without_preload_content` suffix
- remove `_check_type` and `_return_http_data_only`
- replace `_from` with `var_from` for filters
- replace OpenApiModel by BaseModel as OpenApiModel is no longer generated
- add pydantic as a test-requirement
- fix _create_export: the response is no longer a dict
- fix SimpleMetric: it needs to use the Core identifier types

JIRA: CQ-1614
risk: low


no23reason:
Closing as not worth the additional headache now.

PR #1167, opened 2025-10-02
feat(gooddata-pipelines): forbid extra fields when validating provisioning input

janmatzek:


gooddata-pipelines/gooddata_pipelines/provisioning/utils/utils.py
```diff
@@ -42,23 +43,31 @@ def get_attrs(
                         ),
                     }
                 )
+            elif attrs.has(type(context_object)):
+                for key, value in attrs.asdict(
+                    cast(attrs.AttrsInstance, context_object)
```
hkad98, line 48: Nitpick: consider wrapping cast in some try-expect block

hkad98 (approved):

PR #1563, opened 2026-04-20
feat(gooddata-sdk): handle datetime granularity and support bytes limits in arrow fetch

Martozar:
* Add support for datetime granularity handling using `AttributeConverterStore` for fetching data using Arrow. 
* Unify Arrow/AFM API by adding `max_bytes` parameter to Arrow table fetch.
* Improve logging 

risk: low

### DIFF
--- a/packages/gooddata-pandas/src/gooddata_pandas/arrow_convertor.py
+++ b/packages/gooddata-pandas/src/gooddata_pandas/arrow_convertor.py
@@ -5,6 +5,7 @@ from typing import Callable
 
 import orjson
 import pandas
+from gooddata_sdk.type_converter import AttributeConverterStore
 
 from gooddata_pandas.arrow_types import TypesMapper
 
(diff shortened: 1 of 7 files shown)


packages/gooddata-pandas/src/gooddata_pandas/arrow_convertor.py
```diff
@@ -186,10 +250,15 @@ def _build_inline_index(
...
-            if field.name.startswith(_COL_TOTAL_REF_PREFIX):
-                total_ref_vals = table.column(field.name).to_pylist()
-                break
+        total_ref_cols = [f.name for f in table.schema if f.name.startswith(_COL_TOTAL_REF_PREFIX)]
+        if total_ref_cols:
+            if len(total_ref_cols) > 1:
+                logger.warning(
+                    "Arrow table has %d __total_ref* columns; only %r is used for aggregation names.",
+                    len(total_ref_cols),
+                    total_ref_cols[0],
```
lupko, line 259: i wonder, did this actually happen? multiple total ref columns means something is very hosed on the backend...

being defensive is fine, no problem here but if you run into this warning then perhaps there's a more serious problem in xtab impl?

packages/gooddata-sdk/src/gooddata_sdk/compute/model/execution.py
```diff
@@ -408,9 +413,19 @@ def read_result_arrow(self) -> pyarrow.Table:
             _return_http_data_only=True,
         )
         try:
-            return _ipc.open_stream(response).read_all()
+            # Read the full HTTP response body before releasing the connection.
+            # pyarrow's IPC stream reader stops at the Arrow EOS marker and may leave
+            # the HTTP chunked-encoding terminator unread, which puts the connection in
+            # Request-sent state and breaks HTTP keep-alive reuse.
+            data = response.read()
```
lupko: i'm not 100% sure about this. it's worth investigating the implication on memory usage & allocations.

before: arrow reads stream, incrementally builds the table chunk by chunk.
now: likely: first read data for all chunks, then build table. this may (i'm not 100% certain though) to bigger spikes in RSS.

imho its worth investigating / measuring. could be there is no impact due to some technicalities in the underlying arrow code. could be that the load now needs ~2x more memory.

packages/gooddata-pandas/src/gooddata_pandas/arrow_convertor.py
Martozar: No, it didn't happen yet, just a defensive warning "what if", so as you say, if it happens in the future, we should react on this. Original implementation silently picked the first column with the prefix.

packages/gooddata-sdk/src/gooddata_sdk/compute/model/execution.py
Martozar: I did some benchmarks, there was a bit of a memory usage increase, but it wasn't that brutal (can run tests again to see the exact numbers). In any case, from my other measures, Arrow path has less memory consumption than original AFM, so it's kind of a trade-off.

However, I did run into the issue with the  broken connection, and this was a fix for it.

lupko: ok. i wonder though, isn't there a subtler way to address it? 

let's say: try to read stream as before (passing response to Arrow) -> the happy path, ideal memory footprint. then if error happens (except), do response.read() to read everything that is remaining, and then finally close?

or does arrow take ownership of the stream and closes it 'incorrectly' before reading everything or something like that? so exception handler cannot do the cleanup anymore?

lupko (approved):

PR #1580, opened 2026-04-23
feat(gooddata-pipelines): support tags, description and merge into existing LDM

benkeanna:

### DIFF
--- a/docs/content/en/latest/pipelines/ldm_extension/_index.md
+++ b/docs/content/en/latest/pipelines/ldm_extension/_index.md
@@ -45,6 +45,8 @@ The custom dataset represents a new dataset appended to the child LDM. It is def
 | dataset_reference_source_column_data_type | [ColumnDataType](#columndatatype) | Column data type. |
 | workspace_data_filter_id | string | ID of the workspace data filter to use. |
 | workspace_data_filter_column_name | string | Name of the column in custom dataset used for filtering. |
+| dataset_description | string \| None | Optional declarative description on the custom dataset. |
+| dataset_tags | string[] \| None | Optional tag list; when omitted, defaults to a single tag derived from the dataset display name. |
 
 #### Validity constraints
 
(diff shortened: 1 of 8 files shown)


docs/content/en/latest/pipelines/ldm_extension/_index.md
```diff
@@ -128,6 +132,23 @@ ldm_extension_manager.process(
 
 ```
 
+### Merging into an existing child workspace LDM
+
+By default, `process` uploads only the declarative fragment built from your custom dataset definitions. For child workspaces that already inherit an LDM, you typically need to **merge** custom datasets and date instances into the current model instead of replacing it.
```
janmatzek: This confused me a little bit about what to expect from the functionality. Would perhaps something like

```
By default, `process` overwrites the entire everything in child workspace LDM based on your inputs. By setting `merge_into_existing_ldm=True` you can change this behavior to append your inputs to existing child workspace LDM extensions. [...]
```
make it a bit clearer?

packages/gooddata-pipelines/tests/test_ldm_extension/test_ldm_extension_manager.py
```diff
@@ -1,194 +1,108 @@
...
     )
+    manager = _bare_manager(sdk_mock)
 
-    # Should print "Workspace workspace_1 LDM updated." and not revert
-    manager._process_with_relations_check(validated_data)
-    manager._sdk.catalog_workspace_content.put_declarative_ldm.assert_called_once()
-    manager._sdk.catalog_workspace.put_declarative_workspace.assert_not_called()
-
-
-def test_relations_check_failure_and_revert(
```
janmatzek, line 94: is there a reason to remove these tests?

packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/input_processor.py
```diff
@@ -287,3 +307,59 @@ def datasets_to_ldm(
...
+        management_tag: str | None = None,
+    ) -> CatalogDeclarativeModel:
+        """Merge datasets produced from ``custom_datasets`` into an existing declarative LDM.
+
+        Custom datasets and date instances that share an ``id`` with the fragment replace
+        their previous definitions. When ``remove_managed_datasets_missing_from_input`` is
+        set, datasets that carry ``management_tag`` but are absent from the incoming
+        fragment are removed first (typical for tooling-owned extension datasets).
+
+        Inherited LDM objects that are not touched by the fragment remain unchanged.
```
janmatzek: I think you cannot touch the inherited LDM objects anyway. Is this more about the preexisting extensions not being touched by the merge?

janmatzek (changes requested):

janmatzek (approved):

PR #1608, opened 2026-05-12
feat(gooddata-pipelines): support composite key references on parent datasets

benkeanna:
## Summary

Adds a new `ParentDatasetReference` type and a new optional `parent_dataset_references: list[ParentDatasetReference]` field on `CustomDatasetDefinition`, allowing callers to express composite-key joins to the parent dataset (e.g. 2-column foreign key) — a shape the underlying `gooddata_sdk` already supports via `list[CatalogDeclarativeReferenceSource]` but the wrapper currently restricts to a single column.

Driven by a real need on the MIC BCA tooling: BCA datasets reference parent dim datasets that can have composite primary keys, and the existing wrapper API couldn't express that.

### Backward compatibility

- Legacy fields `parent_dataset_reference_attribute_id`, `dataset_reference_source_column`, `dataset_reference_source_column_data_type` remain accepted but are now optional and marked `deprecated=True` (Pydantic emits a `DeprecationWarning` on access).
- A `check_reference_form_exclusive` validator rejects mixing the legacy fields with the new `parent_dataset_references` — chosen over silent precedence to avoid surprising callers later when one path is ignored.
- The processor uses `parent_dataset_references` when set, otherwise falls back to the legacy form. Existing single-column callers see no behavior change.

## Test plan

- [x] `pytest tests/` — 189 passing (composite-only branch).
- [x] `ruff check src/ tests/` — clean.
- [ ] Validate behavior end-to-end against a real workspace with a composite-key parent dim.

### DIFF
--- a/docs/content/en/latest/pipelines/ldm_extension/_index.md
+++ b/docs/content/en/latest/pipelines/ldm_extension/_index.md
@@ -40,9 +40,10 @@ The custom dataset represents a new dataset appended to the child LDM. It is def
 | dataset_source_table | string | Name of the table in the Physical Data Model. |
 | dataset_source_sql | string \| None | SQL query defining the dataset. |
 | parent_dataset_reference | string \| None | ID of the parent dataset to which the custom one will be connected. |
-| parent_dataset_reference_attribute_id | string | ID of the attribute used for creating the relationship in the parent dataset. |
-| dataset_reference_source_column | string | Name of the column used for creating the relationship in the custom dataset. |
-| dataset_reference_source_column_data_type | [ColumnDataType](#columndatatype) | Column data type. |
+| parent_dataset_reference_attribute_id | string \| None | **Deprecated** — use `parent_dataset_references` instead. |
+| dataset_reference_source_column | string \| None | **Deprecated** — use `parent_dataset_references` instead. |
+| dataset_reference_source_column_data_type | [ColumnDataType](#columndatatype) \| None | **Deprecated** — use `parent_dataset_references` instead. |
+| parent_dataset_references | [ParentDatasetReference](#parentdatasetreference)[] \| None | List of references to the parent dataset. |
 | workspace_data_filter_id | string | ID of the workspace data filter to use. |
 | workspace_data_filter_column_name | string | Name of the column in custom dataset used for filtering. |
 | dataset_description | string \| None | Optional declarative description on the custom dataset. |
@@ -51,6 +52,18 @@ The custom dataset represents a new dataset appended to the child LDM. It is def
 #### Validity constraints
 
 Either `dataset_source_table` or `dataset_source_sql` must be specified with a truthy value, but not both. An exception will be raised if both parameters are falsy or if both have truthy values.
+
+`parent_dataset_references` must contain at least one entry.
+
+#### ParentDatasetReference
+
+Bundles one column of a (possibly composite) join to the parent dataset. Pass a list of these on `CustomDatasetDefinition.parent_dataset_references`, one entry per join column.
+
+| name | type | description |
+|------|------|-------------|
+| attribute_id | string | ID of the attribute on the parent dataset that this column joins to. |
+| source_column | string | Name of the column on this dataset used to join to the parent. |
+| data_type | [ColumnDataType](#columndatatype) | Data type of the source column. |
 
 ### Custom Field Definitions
 
@@ -162,6 +175,7 @@ from gooddata_pipelines import (
     CustomFieldDefinition,
     CustomFieldType,
     LdmExtensionManager,
+    ParentDatasetReference,
 )
 
 import logging
@@ -188,9 +202,13 @@ custom_dataset_definitions = [
         dataset_source_table="products_custom",
         dataset_source_sql=None,
         parent_dataset_reference="products",
-        parent_dataset_reference_attribute_id="products.product_id",
-        dataset_reference_source_column="product_id",
-        dataset_reference_source_column_data_type=ColumnDataType.INT,
+        parent_dataset_references=[
+            ParentDatasetReference(
+                attribute_id="products.product_id",
+                source_column="product_id",
+                data_type=ColumnDataType.INT,
+            ),
+        ],
         workspace_data_filter_id="wdf_id",
         workspace_data_filter_column_name="wdf_column",
     )
--- a/packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/input_processor.py
+++ b/packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/input_processor.py
@@ -155,6 +155,46 @@ class LdmExtensionDataProcessor:
         )
 
     @staticmethod
+    def _build_parent_reference_sources(
+        definition: CustomDatasetDefinition,
+    ) -> list[CatalogDeclarativeReferenceSource]:
+        """Build the reference sources from either the new list or the legacy triple."""
+        if definition.parent_dataset_references:
+            return [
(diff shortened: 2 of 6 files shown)


packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/models/custom_data_object.py
```diff
@@ -71,9 +101,34 @@ class CustomDatasetDefinition(BaseModel):
     dataset_source_table: str | None
     dataset_source_sql: str | None
     parent_dataset_reference: str
-    parent_dataset_reference_attribute_id: str
-    dataset_reference_source_column: str
-    dataset_reference_source_column_data_type: ColumnDataType
+    parent_dataset_reference_attribute_id: str | None = Field(
+        default=None,
+        deprecated=(
+            "Use `parent_dataset_references` for richer (composite-key) joins. "
```
janmatzek: I'd leave out the "for richer ... joins" part in the deprecation messages.

packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/models/custom_data_object.py
```diff
@@ -71,9 +101,34 @@ class CustomDatasetDefinition(BaseModel):
...
+            "This field will be removed in a future release."
+        ),
+    )
+    parent_dataset_references: list[ParentDatasetReference] | None = Field(
+        default=None,
+        description=(
+            "Composite-key reference to the parent dataset. When provided and "
+            "non-empty, supersedes the legacy single-column reference fields."
```
janmatzek: I'd go with something like "List of references to parent datasets." or something This phrasing assumes that join on multiple keys is the default use case (llm is going off the conversation context) but for the most users, joining on one field will be more than enough.

packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/models/custom_data_object.py
```diff
@@ -98,6 +153,29 @@ def check_source(self) -> "CustomDatasetDefinition":
             )
         return self
 
+    @model_validator(mode="after")
+    def check_reference_form_exclusive(self) -> "CustomDatasetDefinition":
+        """Reject mixing the legacy single-column fields with ``parent_dataset_references``.
+
+        Forcing callers to pick one form prevents silent precedence surprises:
+        without this check, setting both would quietly use the new list and
+        ignore the legacy values, which is easy to miss when debugging.
```
janmatzek: I'd delete this - it sounds more like AI reasoning than a useful docstring (It tries to justify why the code is present, rather than explain what the code is doing)

packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/input_processor.py
```diff
@@ -154,6 +154,49 @@ def _date_ref_from_field(
...
+    def _build_parent_reference_sources(
+        definition: CustomDatasetDefinition,
+    ) -> list[CatalogDeclarativeReferenceSource]:
+        """Resolve the list of parent reference sources.
+
+        Precedence:
+        * If ``parent_dataset_references`` is set and non-empty, use it as-is.
+        * Otherwise fall back to the legacy single-column fields wrapped in a
+          one-element list. Missing legacy fields yield an empty list, which
+          will be rejected downstream by the GoodData API.
```
janmatzek: actually, would it not be better to fail fast instead of waiting for a an API call to fail?

packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/input_processor.py
```diff
@@ -253,6 +296,13 @@ def datasets_to_ldm(
             # Get the data source info
             dataset_source_table_id, dataset_sql = self._get_sources(dataset)
 
+            # Build the parent reference source list. The composite-friendly
+            # `parent_dataset_references` list takes precedence when set and
+            # non-empty; otherwise fall back to the legacy single-column fields.
```
janmatzek: this is just repeating the docstring of the function that is called. This only needs to exist in one place.

janmatzek (changes requested):

docs/content/en/latest/pipelines/ldm_extension/_index.md
```diff
@@ -40,9 +40,10 @@ The custom dataset represents a new dataset appended to the child LDM. It is def
...
-| parent_dataset_reference_attribute_id | string | ID of the attribute used for creating the relationship in the parent dataset. |
-| dataset_reference_source_column | string | Name of the column used for creating the relationship in the custom dataset. |
-| dataset_reference_source_column_data_type | [ColumnDataType](#columndatatype) | Column data type. |
+| parent_dataset_reference_attribute_id | string \| None | **Deprecated** — single-column reference to the parent attribute. Use `parent_dataset_references` instead. |
+| dataset_reference_source_column | string \| None | **Deprecated** — single-column name on the custom dataset. Use `parent_dataset_references` instead. |
+| dataset_reference_source_column_data_type | [ColumnDataType](#columndatatype) \| None | **Deprecated** — column data type for the single-column reference. Use `parent_dataset_references` instead. |
+| parent_dataset_references | [ParentDatasetReference](#parentdatasetreference)[] \| None | Composite-key reference to the parent dataset (one entry per join column). When set, supersedes the three legacy single-column fields above. |
```
janmatzek: I'd consider using just
```
**Deprecated** — use `parent_dataset_references` instead.
```

docs/content/en/latest/pipelines/ldm_extension/_index.md
```diff
@@ -52,6 +53,18 @@ The custom dataset represents a new dataset appended to the child LDM. It is def
 
 Either `dataset_source_table` or `dataset_source_sql` must be specified with a truthy value, but not both. An exception will be raised if both parameters are falsy or if both have truthy values.
 
+The parent-dataset reference can be expressed via either the three legacy fields (`parent_dataset_reference_attribute_id`, `dataset_reference_source_column`, `dataset_reference_source_column_data_type`) or the new `parent_dataset_references` list — never both. Mixing the two forms raises a `ValidationError`. New code should prefer `parent_dataset_references`; the legacy fields will be removed in a future release.
```
janmatzek: I'd just lead the user straight to the new field. No need to mention the legacy fields here

docs/content/en/latest/pipelines/ldm_extension/_index.md
janmatzek: I think the reference to the composite key is "too much information" here.

packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/input_processor.py
```diff
@@ -154,6 +154,40 @@ def _date_ref_from_field(
...
+                    ),
+                )
+                for ref in definition.parent_dataset_references
+            ]
+
+        # `check_reference_form` on the model guarantees all three legacy
+        # fields are set when `parent_dataset_references` is empty.
+        assert definition.parent_dataset_reference_attribute_id is not None
+        assert definition.dataset_reference_source_column is not None
+        assert definition.dataset_reference_source_column_data_type is not None
```
janmatzek: Maybe it would be better to explicitly throw an exception when any of these is None? I know it should be unreachable because of the validators, but still.

janmatzek (dismissed):

janmatzek (approved):

PR #1609, opened 2026-05-12
feat(gooddata-pipelines): make workspace data filter fields optional

benkeanna:
## Summary

Makes `workspace_data_filter_id` and `workspace_data_filter_column_name` on `CustomDatasetDefinition` optional (`str | None = None`). Previously both were required, which forced callers to provide values even for datasets that don't participate in a workspace data filter — a shape the underlying `gooddata_sdk` already supports (`workspace_data_filter_columns` and `workspace_data_filter_references` on `CatalogDeclarativeDataset` are both `list[...] | None = None`).

Driven by a real need on the MIC BCA tooling: BCA datasets per spec don't have WDF, and the wrapper API was blocking that legitimate use case.

### Backward compatibility

- Existing callers passing both fields keep working byte-for-byte.
- A `check_wdf_pair` validator enforces both-or-neither: setting only one of the two now raises a clear `ValidationError` instead of producing a malformed declarative dataset.
- The processor only emits `workspace_data_filter_columns` / `workspace_data_filter_references` when both fields are non-None; otherwise the SDK receives `None` (which it accepts).

## Test plan

- [x] `pytest tests/` — 186 passing.
- [x] `ruff check src/ tests/` — clean.
- [ ] Validate end-to-end with a real workspace that has a dataset without WDF.

### DIFF
--- a/packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/input_processor.py
+++ b/packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/input_processor.py
@@ -297,6 +297,35 @@ class LdmExtensionDataProcessor:
                 dataset.definition
             )
 
+            wdf_columns: (
+                list[CatalogDeclarativeWorkspaceDataFilterColumn] | None
+            ) = None
+            wdf_references: (
+                list[CatalogDeclarativeWorkspaceDataFilterReferences] | None
+            ) = None
+            wdf_id = dataset.definition.workspace_data_filter_id
+            wdf_column_name = (
+                dataset.definition.workspace_data_filter_column_name
+            )
+            # `check_wdf_pair` on the model guarantees both fields are set
+            # together or both omitted.
+            if wdf_id is not None and wdf_column_name is not None:
+                wdf_columns = [
+                    CatalogDeclarativeWorkspaceDataFilterColumn(
+                        name=wdf_column_name,
+                        data_type=ColumnDataType.STRING.value,
+                    )
+                ]
+                wdf_references = [
+                    CatalogDeclarativeWorkspaceDataFilterReferences(
+                        filter_id=CatalogDatasetWorkspaceDataFilterIdentifier(
+                            id=wdf_id
+                        ),
+                        filter_column=wdf_column_name,
+                        filter_column_data_type=ColumnDataType.STRING.value,
+                    )
+                ]
+
             # Construct the declarative dataset object and append it to the list.
             declarative_datasets.append(
                 CatalogDeclarativeDataset(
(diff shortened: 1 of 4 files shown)


packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/models/custom_data_object.py
```diff
@@ -62,7 +62,12 @@ def check_ids_not_equal(self) -> "CustomFieldDefinition":
 
 
 class CustomDatasetDefinition(BaseModel):
-    """Input model for custom dataset definition."""
+    """Input model for custom dataset definition.
+
+    Workspace data filter fields are optional. Both must be set together or
+    both left unset; when set, a single-column WDF binding is emitted on the
+    declarative dataset.
```
janmatzek: this is obvious from the validator

packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/input_processor.py
```diff
@@ -253,6 +253,34 @@ def datasets_to_ldm(
             # Get the data source info
             dataset_source_table_id, dataset_sql = self._get_sources(dataset)
 
+            # Workspace data filter fields are optional and must be set together
+            # (validated on the input model). Emit columns/references only when
+            # both are provided.
```
janmatzek: This is duplicating the docstring on the validator. Not sure if it is necessary here

packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/input_processor.py
```diff
@@ -283,21 +311,8 @@ def datasets_to_ldm(
...
-                        CatalogDeclarativeWorkspaceDataFilterReferences(
-                            filter_id=CatalogDatasetWorkspaceDataFilterIdentifier(
-                                id=dataset.definition.workspace_data_filter_id
-                            ),
-                            filter_column=dataset.definition.workspace_data_filter_column_name,
-                            filter_column_data_type=ColumnDataType.STRING.value,
-                        )
-                    ],
+                    workspace_data_filter_columns=wdf_columns or None,
+                    workspace_data_filter_references=wdf_references or None,
```
janmatzek: What would happen if we passed the empty lists here?

packages/gooddata-pipelines/src/gooddata_pipelines/ldm_extension/input_processor.py
```diff
@@ -253,6 +253,36 @@ def datasets_to_ldm(
...
+                list[CatalogDeclarativeWorkspaceDataFilterReferences] | None
+            ) = None
+            wdf_id = dataset.definition.workspace_data_filter_id
+            wdf_column_name = (
+                dataset.definition.workspace_data_filter_column_name
+            )
+            if wdf_id is not None:
+                # `check_wdf_pair` on the model guarantees the column name is
+                # set whenever the id is.
+                assert wdf_column_name is not None
```
janmatzek: maybe you can add the `wdf_column_name` to the condition check above?

janmatzek (dismissed):

janmatzek (approved):

PR #1657, opened 2026-06-15
refactor(eval): move all agentic evaluation logic into gooddata_eval SDK

zdenekmusil-gd:
- Add agentic runners for metric_skill, alert_skill, search_tool, general_question, guardrail, and conversation test kinds
- agentic_search pass_at_k requires only tool_selected (matches original Tavern behavior; tool_correctness is a Langfuse quality metric only)
- Expose evaluate_agentic_* functions for use by Tavern thin shims
- Update uv.lock

JIRA: GDAI-1830
risk: nonprod

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
+++ b/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
@@ -1,0 +1,236 @@
+# (C) 2026 GoodData Corporation. All rights reserved.
+"""Agentic evaluation runner for gd-eval CLI — handles multi-turn agentic test kinds."""
+
+from __future__ import annotations
+
+import time
+from typing import Any, TypedDict
+
+from gooddata_eval.core.agentic._langfuse import HttpxLangfuseClient, make_langfuse_client
+from gooddata_eval.core.agentic.alert_skill import evaluate_agentic_alert_skill
+from gooddata_eval.core.agentic.conversation import ConversationFixture, evaluate_agentic_conversation
+from gooddata_eval.core.agentic.general_question import evaluate_agentic_general_question
+from gooddata_eval.core.agentic.guardrail import evaluate_agentic_guardrail
+from gooddata_eval.core.agentic.metric_skill import evaluate_agentic_metric_skill
+from gooddata_eval.core.agentic.search_tool import evaluate_agentic_search_tool
+from gooddata_eval.core.agentic.visualization import evaluate_agentic_visualization
+from gooddata_eval.core.models import CreatedVisualization, DatasetItem
+from gooddata_eval.core.runner import EvalReport, ItemReport
+
+_LfKw = TypedDict(
+    "_LfKw",
+    {
+        "langfuse": Any,
+        "dataset_item_id": str,
+        "dataset_name": str,
+        "run_timestamp": str,
+        "model_version_override": str | None,
+    },
+    total=False,
+)
+
+AGENTIC_TEST_KINDS = frozenset(
+    {
+        "vis_agentic",  # production: expected_output.visualization (single/multi CreatedVisualization)
+        "agentic_visualization",  # experimental: expected_output.expected_outputs (multi-candidate)
+        "agentic_metric_skill",
+        "agentic_alert_skill",
+        "agentic_search",
+        "agentic_general_question",
+        "agentic_guardrail",
+        "agentic_conversation",
+    }
+)
+
+
+def _parse_visualization_expected(expected_output: Any) -> list[CreatedVisualization]:
+    """Parse expected_output into a list of CreatedVisualization candidates.
+
+    Accepts:
+      {"expected_outputs": [{"visualization": {...}}, ...]}  <- agentic fixture format
+      {"visualization": {...}} or {"visualization": [{...}]}  <- single/multi candidate
+      [{"visualization": {...}}, ...]                          <- bare list
+    """
+    if isinstance(expected_output, dict):
+        raw_list = expected_output.get("expected_outputs")
+        if raw_list is not None:
+            return [
+                CreatedVisualization.model_validate(v.get("visualization", v) if isinstance(v, dict) else v)
+                for v in raw_list
+            ]
+        raw_viz = expected_output.get("visualization")
+        if raw_viz is not None:
+            if isinstance(raw_viz, list):
+                return [CreatedVisualization.model_validate(v) for v in raw_viz]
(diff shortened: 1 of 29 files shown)


packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
```diff
@@ -0,0 +1,209 @@
+# (C) 2026 GoodData Corporation. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-GoodData-Enterprise
```
hkad98: What is the purpose of this: SPDX-License-Identifier?

packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
```diff
@@ -0,0 +1,209 @@
...
+    lf_kw = dict(
+        langfuse=langfuse,
+        dataset_item_id=item.id,
+        dataset_name=item.dataset_name,
+        run_timestamp=run_ts,
+        model_version_override=model_version_override,
+    )
+
+    if kind in ("vis_agentic", "agentic_visualization"):
+        from gooddata_eval.core.agentic.visualization import evaluate_agentic_visualization  # noqa: PLC0415
```
hkad98: Is there any reason to ignore PLC0415 and having the import here and not in top level?

packages/gooddata-eval/src/gooddata_eval/cli/main.py
```diff
@@ -62,6 +63,17 @@ def _build_parser() -> argparse.ArgumentParser:
     source = run.add_mutually_exclusive_group(required=True)
     source.add_argument("--dataset", help="Path to a folder of dataset JSON files.")
     source.add_argument("--langfuse-dataset", dest="langfuse_dataset", help="Langfuse dataset name.")
+    run.add_argument(
```
hkad98, line 66: I do not see `visualization` in AGENTIC_TEST_KINDS.

packages/gooddata-eval/src/gooddata_eval/core/agentic/_langfuse.py
```diff
@@ -0,0 +1,378 @@
...
+    def create(
+        self,
+        run_name: str,
+        dataset_item_id: str,
+        trace_id: str,
+        metadata: dict | None = None,
+        run_description: str = "",
+    ) -> None:
+        self._client.post(
+            "/api/public/dataset-run-items",
```
hkad98, line 70: Is not there support in official Langfuse SDK for this?

packages/gooddata-eval/src/gooddata_eval/core/agentic/_langfuse.py
```diff
@@ -0,0 +1,378 @@
...
+    """Return model version: explicit override > workspace active LLM provider."""
+    if override:
+        return override
+    try:
+        import httpx  # noqa: PLC0415
+        resp = httpx.get(
+            f"{host}/api/v1/entities/workspaces/{workspace_id}/workspaceSettings/activeLlmProvider",
+            headers={"Accept": "application/json", "Authorization": f"Bearer {token}"},
+            timeout=5,
+        )
```
hkad98: We should use `gooddata-sdk` for this.

packages/gooddata-eval/src/gooddata_eval/core/agentic/alert_skill.py
```diff
@@ -0,0 +1,451 @@
+# (C) 2026 GoodData Corporation. All rights reserved.
```
hkad98, line 1: There should be also representation for alerts in gooddata-sdk. Please have a look if we could reuse it. gooddata-sdk should be source of the truth for oapi spec and it schemas.

packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
zdenekmusil-gd: Yes — these are deferred intentionally. The openai package is an optional [llm-judge] extra; importing it at module level would crash the CLI for users who haven't installed it. The other imports (evaluate_agentic_*) are deferred to avoid pulling in all evaluation dependencies on every CLI invocation.

packages/gooddata-eval/src/gooddata_eval/core/agentic/_langfuse.py
zdenekmusil-gd: Yes, the official SDK has create_dataset_run_item, but we can't use it — the Langfuse Python SDK relies on Pydantic v1 shims that break at import time on Python 3.14, which gooddata-eval targets.

packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
hkad98: Proposal what if we checked for openai dependency and then import those in top-level? The motivation for that is to have readable code. Pulling those imports in top-level is not expensive.

packages/gooddata-eval/src/gooddata_eval/core/agentic/_catalog.py
```diff
@@ -0,0 +1,49 @@
+# (C) 2026 GoodData Corporation
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+
+@dataclass
+class CatalogMetricAlert:
```
hkad98, line 8: This should inherit for base class otherwise you cannot use this in API calls. 

Base class: https://github.com/gooddata/gooddata-python-sdk/blob/95b2cea31b84a3d828c2f8a5d69758ab5beb48a5/gooddata-api-client/gooddata_api_client/model/automation_alert.py#L40

See for example: https://github.com/gooddata/gooddata-python-sdk/blob/95b2cea31b84a3d828c2f8a5d69758ab5beb48a5/packages/gooddata-sdk/src/gooddata_sdk/catalog/workspace/declarative_model/workspace/automation.py#L55

packages/gooddata-eval/src/gooddata_eval/core/agentic/_langfuse.py
hkad98: Is there a fix for that? Do we have a GH issue that we could link and fix later?

zdenekmusil-gd: https://gooddata.atlassian.net/browse/GDAI-1913

hkad98 (approved):

PR #1667, opened 2026-06-19
zmu/gdai 1929 genai sse transient retry

zdenekmusil-gd:
Classify SSE error events as ChatError (non-retryable, still a RuntimeError) vs TransientChatError (status 429/502/503/504 or METADATA_SYNC_IN_PROGRESS, detected before JSON parse). Wrap ChatClient.send_message and create_conversation in a bounded exponential-backoff retry (5 retries, 5/10/20/40/60s, ~2min cap, stdlib only) so a staging metadata-sync queue peak no longer fails the E2E LLM tests. No new dependency.

JIRA: GDAI-1929
risk: nonprod

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/chat/sse_client.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/chat/sse_client.py
@@ -14,14 +14,87 @@ protocol, not on this class.
 """
 
 import json
+import logging
+import os
+import time
 from dataclasses import dataclass, field
-from typing import Any, Iterable
+from typing import Any, Callable, Iterable, TypeVar
 
 import httpx
 
 from gooddata_eval.core.models import ChatResult, DatasetItem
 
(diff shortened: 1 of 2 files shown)


packages/gooddata-eval/src/gooddata_eval/core/chat/sse_client.py
```diff
@@ -14,15 +14,73 @@
...
+
+
+class TransientChatError(ChatError):
+    """Retryable transient error: gen-ai temporarily unavailable or still syncing metadata."""
+
+
+_MAX_RETRIES = 5
+_INITIAL_BACKOFF_S = 5.0
+_BACKOFF_FACTOR = 2.0
+_MAX_BACKOFF_S = 60.0
```
hkad98: Consider making this configurable via env var.

hkad98 (approved):

PR #1675, opened 2026-06-27
fix(eval): fix search_tool correctness always scoring 0%

Tomkess:
## Summary

The `tool_correctness` metric in `search_tool` evaluations was always `False` (0%), making the dashboard metric useless.

**Root cause:** `_args_match` checked two fields that never matched real model calls:
- `emit_widget` — this parameter was renamed to `user_requested_search` in the tool schema; the model never sends `emit_widget`, so `actual_args.get("emit_widget")` was always `None` vs the expected `false`
- `limit` — declared as optional (`int | None`) in the schema; the model omits it and relies on the server default, while fixtures hardcode `10`

**Fix:** Only check `keywords` (case-insensitive) and `object_types` — the two fields that determine whether the search was semantically correct. `limit` and the widget display flag are not part of search correctness.

## Test Plan

- [ ] Run `uv run pytest tests/ -x -q` in `packages/gooddata-eval`
- [ ] Re-run search_tool eval items and verify `tool_correctness` now reflects actual model behaviour (models that call with the right keywords and object types should score True)

## Risk

Low — single function change in the evaluator, no effect on eval runner or production code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved matching of search tool calls to judge correctness using the most relevant inputs.
  * Search keywords are now compared case-insensitively, and object types are normalized for more reliable matching.
  * Search options such as limit and widget emission are no longer required to match exactly, reducing false mismatches.
  * Malformed or non-string keyword/object type inputs are handled defensively, preventing matching-time errors.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
@@ -6,9 +6,12 @@ from gooddata_eval.core.models import ChatResult, DatasetItem
 
 
 def _normalize_str_list(value: object, *, lowercase: bool = False) -> list[str]:
-    # Arguments come from raw model-emitted JSON, so a malformed tool call may
-    # contain non-string entries. Drop them defensively so bad input scores as a
-    # mismatch instead of raising and aborting the whole evaluation.
+    # Arguments come from raw model-emitted JSON. The search_objects schema
+    # declares keywords/object_types as list[str], but a malformed tool call may
+    # send a non-list or non-string entries. Drop the offending entries defensively
+    # so bad input can't raise (.lower()/sorted() on a non-str) and abort the whole
+    # evaluation run; a non-list collapses to [] and the surviving strings are
+    # still compared normally.
     if not isinstance(value, list):
         return []
     items = [item for item in value if isinstance(item, str)]
@@ -23,6 +26,9 @@ def _args_match(actual_args: dict, expected_args: dict) -> bool:
     expected_kw = _normalize_str_list(expected_args.get("keywords"), lowercase=True)
     if actual_kw != expected_kw:
         return False
+    # object_types is compared case-sensitively (no lowercase=True): they are
+    # controlled ObjectType StrEnum values the model emits verbatim ("metric",
+    # "dashboard"), so a case mismatch is a genuine error, not a formatting quirk.
     return _normalize_str_list(actual_args.get("object_types")) == _normalize_str_list(
         expected_args.get("object_types")
     )
--- a/packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
@@ -6,13 +6,14 @@ from gooddata_eval.core.models import ChatResult, DatasetItem
 
 
 def _args_match(actual_args: dict, expected_args: dict) -> bool:
-    if sorted(actual_args.get("keywords") or []) != sorted(expected_args.get("keywords") or []):
+    # Only keywords and object_types determine semantic correctness.
+    # limit is optional with a server-side default; emit_widget was renamed to
+    # user_requested_search in the tool schema — neither affects search quality.
+    actual_kw = sorted(k.lower() for k in (actual_args.get("keywords") or []))
+    expected_kw = sorted(k.lower() for k in (expected_args.get("keywords") or []))
+    if actual_kw != expected_kw:
         return False
-    if sorted(actual_args.get("object_types") or []) != sorted(expected_args.get("object_types") or []):
-        return False
-    if actual_args.get("limit") != expected_args.get("limit"):
-        return False
-    return actual_args.get("emit_widget") == expected_args.get("emit_widget")
+    return sorted(actual_args.get("object_types") or []) == sorted(expected_args.get("object_types") or [])
 
 
 class SearchToolEvaluator:
--- a/packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
@@ -5,15 +5,27 @@ from gooddata_eval.core.evaluators.base import ItemEvaluation
 from gooddata_eval.core.models import ChatResult, DatasetItem
 
 
+def _normalize_str_list(value: object, *, lowercase: bool = False) -> list[str]:
+    # Arguments come from raw model-emitted JSON, so a malformed tool call may
+    # contain non-string entries. Drop them defensively so bad input scores as a
+    # mismatch instead of raising and aborting the whole evaluation.
+    if not isinstance(value, list):
+        return []
+    items = [item for item in value if isinstance(item, str)]
+    return sorted(item.lower() if lowercase else item for item in items)
+
+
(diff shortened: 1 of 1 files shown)


packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
```diff
@@ -6,13 +6,14 @@
...
+    actual_kw = sorted(k.lower() for k in (actual_args.get("keywords") or []))
+    expected_kw = sorted(k.lower() for k in (expected_args.get("keywords") or []))
+    if actual_kw != expected_kw:
         return False
-    if sorted(actual_args.get("object_types") or []) != sorted(expected_args.get("object_types") or []):
-        return False
-    if actual_args.get("limit") != expected_args.get("limit"):
-        return False
-    return actual_args.get("emit_widget") == expected_args.get("emit_widget")
+    return sorted(actual_args.get("object_types") or []) == sorted(expected_args.get("object_types") or [])
```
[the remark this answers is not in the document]
Tomkess: Good catch — applied in f8476798. Added a `_normalize_str_list` helper that drops non-string entries before lowercasing/sorting, so malformed tool-call JSON (e.g. numeric keywords or mixed `object_types`) now scores `tool_correctness=False` instead of raising and aborting the eval run. This matters here since we run evals against small local models that are more prone to emitting malformed tool calls. Verified with the existing search_tool tests (7 passed) plus manual checks of the numeric/mixed-type cases.

packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
```diff
@@ -5,14 +5,27 @@
 from gooddata_eval.core.models import ChatResult, DatasetItem
 
 
+def _normalize_str_list(value: object, *, lowercase: bool = False) -> list[str]:
+    # Arguments come from raw model-emitted JSON, so a malformed tool call may
+    # contain non-string entries. Drop them defensively so bad input scores as a
+    # mismatch instead of raising and aborting the whole evaluation.
+    if not isinstance(value, list):
```
hkad98, line 15: **Question:** `_normalize_str_list` returns `[]` for any non-list input, so `keywords` arriving as a bare string collapses to empty. The sibling agentic evaluator explicitly handles that case — `agentic/search_tool.py:33` compares `keywords` as a string (and its tests pass `{"keywords": "revenue"}`). If `search_objects` can emit `keywords` as a string rather than a list, a correct call scores `tool_correctness=False` here — a false-negative in a new disguise. What does the tool schema actually emit for `keywords`, list or string?

---
Posted by Claude after discussing with @hkad98

packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
```diff
@@ -5,14 +5,27 @@
...
-    if actual_args.get("limit") != expected_args.get("limit"):
+    # Only keywords and object_types determine semantic correctness.
+    # limit is optional with a server-side default; emit_widget was renamed to
+    # user_requested_search in the tool schema — neither affects search quality.
+    actual_kw = _normalize_str_list(actual_args.get("keywords"), lowercase=True)
+    expected_kw = _normalize_str_list(expected_args.get("keywords"), lowercase=True)
+    if actual_kw != expected_kw:
         return False
-    return actual_args.get("emit_widget") == expected_args.get("emit_widget")
+    return _normalize_str_list(actual_args.get("object_types")) == _normalize_str_list(
```
hkad98, line 32: **Nit:** `keywords` is matched case-insensitively (`lowercase=True`) but `object_types` is case-sensitive. Presumably intentional since `object_types` are controlled enums the model emits verbatim (`metric`, `dashboard`) — but a stray `"Metric"` would mismatch. Worth a one-line comment noting the asymmetry is deliberate.

---
Posted by Claude after discussing with @hkad98

packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
```diff
@@ -5,14 +5,27 @@
 from gooddata_eval.core.models import ChatResult, DatasetItem
 
 
+def _normalize_str_list(value: object, *, lowercase: bool = False) -> list[str]:
+    # Arguments come from raw model-emitted JSON, so a malformed tool call may
+    # contain non-string entries. Drop them defensively so bad input scores as a
```
hkad98: **Nit:** The comment says dropping non-string entries makes bad input "score as a mismatch," but `_normalize_str_list(["dashboard", 123])` → `["dashboard"]`, which *matches* expected `["dashboard"]`. It's a match (arguably fine — the meaningful strings are right), not a mismatch. Minor wording tweak.

---
Posted by Claude after discussing with @hkad98

hkad98:
**Question:** This changes scoring behavior but adds no test locking it in. The two existing cases in `test_search_tool_evaluator.py` still pass only because their fixtures keep `limit`/`emit_widget` matching — nothing asserts the actual fix: a call where `limit`/`emit_widget` **differ** but `keywords`+`object_types` match should now score `tool_correctness=True`. Given this metric was silently wrong in production, could we add a regression test for that (plus case-insensitive keywords and the malformed-input path)?

---
Posted by Claude after discussing with @hkad98

packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
Tomkess: Good question — I checked the tool schema. `search_objects` is backed by `VectorSearchParams` in gen-ai ([`vector_search_tool.py`](https://github.com/gooddata/gdc-nas/blob/master/microservices/gen-ai/app/infrastructure/services/tools/vector_search_tool.py)), where both fields are declared as lists:

```python
keywords: list[str] | None = Field(...)
object_types: list[ObjectType] = Field(...)
```

So `keywords` is always emitted as a list (or omitted), never as a bare string — the string-collapse false-negative can't occur for this evaluator. The sibling `agentic/search_tool.py` handles a string `keywords` because its `_tool_correctness` is a *generic* arg-comparison loop over arbitrary `expected_tool_call` keys (and its test fixture happens to pass a string), not because the schema emits strings there. I've added a note to the `_normalize_str_list` comment stating the schema declares `list[str]` so the intent is explicit. No change to matching behavior needed.

packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
Tomkess: Correct, it's deliberate — `object_types` are controlled `ObjectType` `StrEnum` values (`metric`, `dashboard`, …, all lowercase) the model emits verbatim, so a case difference there is a genuine error rather than a formatting quirk. Added a one-line comment noting the asymmetry is intentional (commit 53dc5d2e).

packages/gooddata-eval/src/gooddata_eval/core/evaluators/search_tool.py
Tomkess: Fair point — reworded in 53dc5d2e. The comment now says dropping non-string entries prevents a crash (`.lower()`/`sorted()` on a non-str) and that surviving strings are still compared normally, rather than claiming bad input "scores as a mismatch."

hkad98 (approved):

PR #1683, opened 2026-07-08
fix(eval): treat missing/null alert trigger as ALWAYS default

henrynguyengooddata:
## Summary

`agent_alert_skill` intermittently failed on `trigger_correct=False` for **"Every time"** alert cases whenever the agent created the alert without an explicit trigger. `create_metric_alert` serialises an unset trigger as `trigger: null`, so `actual_args.get("trigger", "ALWAYS")` returned `None` — the intended ALWAYS fallback only applies when the key is *absent*, not when it is present-but-null. Since the product default for an unset trigger **is** `ALWAYS` ("Every time"), this was a false-negative.

Fix: default a missing **or** null trigger to `ALWAYS` in both scoring paths:
- `core/agentic/alert_skill.py::_check_trigger` (used by the agentic runner that CI runs)
- `core/evaluators/alert_skill.py::AlertSkillEvaluator` (registered evaluator)

A genuinely wrong trigger (e.g. `ONCE_PER_INTERVAL` when `ONCE` is expected) still fails — the fix does not mask real model errors.

## Evidence — replayed the exact `actual_args` from failing CI runs

| Case (real args captured from CI) | Expected | Before (`master`) | After (this PR) |
|---|---|:---:|:---:|
| "every time returns > 150", `trigger=None` | pass | ❌ False | ✅ True |
| "returns by brand > 30 last day", `trigger=None` | pass | ❌ False | ✅ True |
| "the **first time** customers < 200", `ONCE_PER_INTERVAL` | fail | ❌ False | ❌ False |

Both scoring paths (`_check_trigger` and `AlertSkillEvaluator`) agree. The first two rows reproduce the exact CI false-negatives and flip to pass; the third (a real model error) correctly stays failing.

## Test plan

- New regression tests:
  - `tests/test_agentic_alert_skill.py` — `_check_trigger`: missing/null trigger → ALWAYS passes; a "One time" expectation still requires an explicit `ONCE`.
  - `tests/test_alert_skill_evaluator.py` — evaluator defaults missing/null trigger to ALWAYS.
- `pytest packages/gooddata-eval/tests/test_agentic_alert_skill.py packages/gooddata-eval/tests/test_alert_skill_evaluator.py` → **17 passed**.

## Downstream

gdc-nas `tests/tavern-e2e/` pins `gooddata-eval[llm-judge]`. After this is released, a follow-up bumps the pin + re-locks `uv.lock` so the fix reaches the nightly agent-evaluation.

## References

- Failing CI (gooddata/gdc-nas): run `28479836676` (2026-06-30), run `28902805843` (2026-07-07)
- Internal tracking: QA-28465

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved alert trigger handling so missing or empty trigger values now behave like the default “Every time” setting.
  * Ensured trigger checks consistently require an explicit one-time trigger when that option is expected.
  * Added coverage for missing and null trigger cases to prevent regressions.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/evaluators/alert_skill.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/evaluators/alert_skill.py
@@ -78,7 +78,9 @@ class AlertSkillEvaluator:
 
         if "Trigger" in expected:
             expected_trigger = _TRIGGER_MAP.get(expected["Trigger"], expected["Trigger"])
-            trigger_correct = args.get("trigger") == expected_trigger
+            # Omitted/null trigger persists as the product default ALWAYS ("Every time").
+            actual_trigger = args.get("trigger") or "ALWAYS"
+            trigger_correct = actual_trigger == expected_trigger
 
         if "Filters" in expected:
             actual_filters = args.get("filters") or []
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/alert_skill.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/alert_skill.py
@@ -78,7 +78,11 @@ def _check_trigger(expected: CatalogMetricAlert, actual_args: dict) -> bool:
     if expected.operator == "ANOMALY":
         return True
     exp_trigger = expected.trigger
-    act_trigger = actual_args.get("trigger", actual_args.get("triggerMode", "ALWAYS"))
+    # A missing OR explicit-null trigger means "unset" -> the product persists the
+    # default ALWAYS ("Every time"). `.get(k, default)` only returns the default when
+    # the key is ABSENT, but create_metric_alert serialises unset params as
+    # `trigger: null`, so chain with `or` to also cover the present-but-None case.
+    act_trigger = actual_args.get("trigger") or actual_args.get("triggerMode") or "ALWAYS"
     if exp_trigger in _ALWAYS_TRIGGER_VALUES:
         return act_trigger in {"ALWAYS", "Every time"}
     act_api = _TRIGGER_DISPLAY_TO_API.get(act_trigger, act_trigger)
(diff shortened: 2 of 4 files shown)


packages/gooddata-eval/src/gooddata_eval/core/evaluators/alert_skill.py
```diff
@@ -78,7 +78,9 @@ def evaluate(self, item: DatasetItem, chat_result: ChatResult) -> ItemEvaluation
 
         if "Trigger" in expected:
             expected_trigger = _TRIGGER_MAP.get(expected["Trigger"], expected["Trigger"])
-            trigger_correct = args.get("trigger") == expected_trigger
+            # Omitted/null trigger persists as the product default ALWAYS ("Every time").
+            actual_trigger = args.get("trigger") or "ALWAYS"
```
hkad98, line 82:
```suggestion
            actual_trigger = args.get("trigger","ALWAYS")
```

henrynguyengooddata: Thanks @hkad98! I used `or "ALWAYS"` deliberately instead of `.get("trigger", "ALWAYS")`. `.get(key, default)` only falls back when the key is **absent**, but gpt-5.2 calls `create_metric_alert` with `"trigger": null` — the key is present with a null value (`parsed_arguments()` is a raw `json.loads`, no schema merge). So `args.get("trigger", "ALWAYS")` returns `None`, and `None == "ALWAYS"` → `trigger_correct=False`, i.e. the exact false-negative this PR fixes would come back.

For the real failing args `{"trigger": None, ...}`:
- `args.get("trigger", "ALWAYS")` → `None` → fails
- `args.get("trigger") or "ALWAYS"` → `"ALWAYS"` → passes

(The original agentic `_check_trigger` already used `.get("trigger", …default…)` and still produced the CI false-negatives for the same reason.) Since no valid trigger value is falsy (`ALWAYS`/`ONCE`/`ONCE_PER_INTERVAL`), `or` safely covers both the null and absent cases. Happy to make it more explicit if you prefer: `raw = args.get("trigger"); actual_trigger = raw if raw is not None else "ALWAYS"`.

hkad98 (approved):

PR #1685, opened 2026-07-10
fix(gooddata-eval): delete metrics created during agentic eval runs

myhoai:
Metric-skill and conversation evaluators create metrics in a shared, persistent eval workspace but never removed them. A leftover metric is detected and reused by a later test (the agent returns `SELECT {existing_id}` instead of full MAQL), so the assertion fails — and it makes some conversation metric-turns produce no create_metric call at all.

Delete each metric a run creates on the way out, keyed on the exact metric_id from the create_metric tool result, in a finally block:
- metric_skill: per-run cleanup in _execute_single_metric_run.
- conversation: collect metric_ids across turns, delete after the conversation completes (deferred so a later turn's $ref still resolves).

Mirrors the existing alert_skill._delete_alert pattern. Parallelism-safe: each run deletes only the id it created, so concurrent tests sharing the workspace never touch each other's objects. No version bump.

JIRA: QA-28379
risk: nonprod

## Summary by CodeRabbit

* **Bug Fixes**
  * Added best-effort cleanup of all metrics created during multi-turn agentic conversations, ensuring each unique metric is deleted from the workspace after the run.
  * Added best-effort cleanup of metrics created during metric-skill evaluations, including consistent deletion across repeated metric runs.
  * Cleanup now occurs even when later turns fail, preventing leftover metrics from affecting subsequent runs.
* **Tests**
  * Added new coverage to verify correct metric deletion requests, deduplication across turns, and cleanup behavior when errors occur mid-run.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
@@ -8,8 +8,10 @@ import re
 from dataclasses import dataclass
 from typing import Literal
 
+from gooddata_sdk import GoodDataSdk
 from pydantic import BaseModel
 
+from gooddata_eval.core.agentic.metric_skill import _delete_metric, _extract_created_metric_ids
 from gooddata_eval.core.chat.sse_client import ChatClient
 from gooddata_eval.core.models import ChatResult, ToolCallEvent
 from gooddata_eval.core.scoring import (
@@ -283,11 +285,16 @@ def run_agentic_conversation(
     replies before the agent produces the expected output.
     """
     client = ChatClient(host=host, token=token, workspace_id=workspace_id)
+    sdk = GoodDataSdk.create(host, token)
     turn_results: list[TurnResult] = []
     turn_outputs: dict[str, dict] = {}
     total_clarification_turns = 0
     conversation_id: str = ""
     owns_conversation = False
+    # Metrics created during this conversation, deleted after it completes so they do
+    # not persist in the (shared) workspace and get reused by a later test. Deferred to
+    # the end — a later turn may $ref a metric an earlier turn created.
+    created_metric_ids: list[str] = []
 
     try:
         if initial_conversation_id is not None:
@@ -335,6 +342,11 @@ def run_agentic_conversation(
                 if metric_data:
                     turn_outputs[turn.turn_id] = metric_data
 
+            # Track every metric created this turn (any turn may create one) for cleanup.
+            for metric_id in _extract_created_metric_ids(all_tool_calls):
+                if metric_id not in created_metric_ids:
+                    created_metric_ids.append(metric_id)
+
             turn_results.append(
                 TurnResult(
                     turn_id=turn.turn_id,
@@ -351,6 +363,8 @@ def run_agentic_conversation(
     finally:
         if owns_conversation and conversation_id:
             client.delete_conversation(conversation_id)
+        for metric_id in created_metric_ids:
+            _delete_metric(sdk, workspace_id, metric_id)
         client.close()
 
     activated_all = {skill for tr in turn_results for skill in tr.activated_skills}
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
@@ -8,6 +8,8 @@ import re
 from dataclasses import dataclass
 from typing import Any
 
+from gooddata_sdk import GoodDataSdk
+
 from gooddata_eval.core.chat.sse_client import ChatClient
 from gooddata_eval.core.models import ToolCallEvent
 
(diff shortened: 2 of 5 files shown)


packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
```diff
@@ -127,6 +127,43 @@ def _extract_metric_result(tool_call_events: list[ToolCallEvent]) -> dict | None
...
+        if tc.function_name != "create_metric" or not tc.result:
+            continue
+        result_data = tc.parsed_result()
+        if not result_data:
+            continue
+        data = result_data.get("data", result_data)
+        metric_id = data.get("metric_id") if isinstance(data, dict) else None
+        if metric_id and metric_id not in metric_ids:
+            metric_ids.append(metric_id)
+    return metric_ids
```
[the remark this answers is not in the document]
myhoai: I'm going to keep the single-ID path in metric_skill and not apply this here — the reasoning is that the metric-skill flow creates at most one metric per run by design (the metric-skill produces a single metric per request, and the run breaks as soon as create_metric is observed).

packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
```diff
@@ -136,41 +153,53 @@ def _is_asking_clarification(text: str) -> bool:
...
-    )
+    try:
+        for _iteration in range(max_iterations):
+            turns += 1
+            chat_result = client.send_message(conversation_id, current_question)
+            candidate = _extract_metric_result(chat_result.tool_call_events or [])
+            if candidate is not None:
+                metric_result = candidate
+                metric_id_to_delete = candidate.get("metric_id")
+                break
```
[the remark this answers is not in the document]
myhoai: I'm going to keep the single-ID path in metric_skill and not apply this here — the reasoning is that the metric-skill flow creates at most one metric per run by design (the metric-skill produces a single metric per request, and the run breaks as soon as create_metric is observed).

packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
```diff
@@ -335,6 +339,13 @@ def run_agentic_conversation(
                 if metric_data:
                     turn_outputs[turn.turn_id] = metric_data
 
+            # Track every metric created this turn (any turn may create one) for cleanup.
+            from gooddata_eval.core.agentic.metric_skill import _extract_created_metric_ids  # noqa: PLC0415
```
hkad98: Is `# noqa: PLC0415` necessary here? Why this cannot be top level import?

packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
```diff
@@ -351,6 +362,10 @@ def run_agentic_conversation(
     finally:
         if owns_conversation and conversation_id:
             client.delete_conversation(conversation_id)
+        from gooddata_eval.core.agentic.metric_skill import _delete_metric  # noqa: PLC0415
```
hkad98: Is # noqa: PLC0415 necessary here? Why this cannot be top level import?

packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
```diff
@@ -127,6 +127,43 @@ def _extract_metric_result(tool_call_events: list[ToolCallEvent]) -> dict | None
...
+    MAQL) and the assertion fails. Deleting the created metric on the way out keeps
+    the workspace clean for the next run. Mirrors ``alert_skill._delete_alert``.
+    """
+    host = str(client._base).split("/api/")[0]
+    url = f"{host}/api/v1/entities/workspaces/{workspace_id}/metrics/{metric_id}"
+    try:
+        response = client._client.delete(url, headers=client._auth)
+        response.raise_for_status()
+    except Exception as exc:
+        print(f"[CLEANUP] Failed to delete metric {metric_id}: {exc}")
```
hkad98: gooddata-eval imports gooddata-sdk – if possible use it. Do not reimplement logic here.

packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
myhoai: updated! use top-level imports

packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
myhoai: used top-level imports

packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
myhoai: fixed! fix for alert skill test also

hkad98 (approved):

PR #1705, opened 2026-08-04
feat(gooddata-sdk): report whether create_or_update created or updated

vondravl:
## What

Adds `UpsertOutcome` (`CREATED` / `UPDATED`) and returns it from all nine `create_or_update*` methods:

| service | method |
| -- | -- |
| `catalog_workspace` | `create_or_update`, `create_or_update_workspace_setting`, `create_or_update_user_data_filter`, `create_or_update_filter_view` |
| `catalog_user` | `create_or_update_user`, `create_or_update_user_group` |
| `catalog_organization` | `create_or_update_jwk`, `create_or_update_export_template` |
| `catalog_data_source` | `create_or_update_data_source` |

## Why

Each of these already does an existence check and branches on it, then discards the answer:

```python
try:
    found_workspace = self.get_workspace(workspace.id)
    ...update...
except NotFoundException:
    ...create...
```

A caller that needs to know which happened — to log accurately, count creations, or skip follow-up work — has to repeat the GET the SDK just made, because `create_entity_*` / `update_entity_*` are reachable only through the private `_entities_api`. This returns information the SDK has already computed.

The concrete case that prompted it: a deploy tool logging `"Workspace %s created successfully"` after every `create_or_update`, reporting a creation on every re-upload of an existing workspace. The only public-API fix is a duplicate `get_workspace()` call.

## Should this be in the SDK at all?

I genuinely don't know, and I'd rather surface the trade-off than argue one side. **Happy to close this if the answer is no** — the workaround (an extra GET in the caller) is cheap.

**Reasons to take it**

- The information exists and is thrown away. No new I/O, no new failure mode.
- It removes a duplicate round-trip from every consumer that currently needs the answer.
- Backward compatible in practice: every in-repo caller (`gooddata-pipelines`, `gooddata-dbt`, `gooddata-eval`) ignores the return value, and adding a return to a `-> None` method breaks no caller or type checker.
- There's recent precedent for surfacing already-computed results — #1694 (*report created metric/automation id from single-turn evaluators*).

**Reasons to reject it**

- **It may be a deliberate API-surface choice.** Void upserts keep "the entity now looks like this" as the whole contract. Returning the branch invites callers to depend on a distinction the HTTP API itself doesn't promise.
- **The outcome is best-effort, not authoritative.** The existence check is not atomic with the write that follows, so a concurrent create/delete can make the reported branch wrong. The enum docstring says so, but a returned value tends to get trusted more than a docstring caveat. If anyone treats it as an audit record, that's a footgun the current `None` doesn't have.
- **It widens the public API by nine signatures plus a new exported symbol**, to serve a fairly narrow need — arguably the caller's own `get_*` probe is the honest place to pay for that knowledge.
- **Nine methods change shape at once.** Doing only the one I needed would be worse (arbitrary inconsistency), but the all-or-nothing framing makes the smallest useful version of this change fairly wide.
- A future true-upsert endpoint (single PUT, no pre-GET) would make the pre-check — and this return value — obsolete or harder to compute.

For what it's worth on intent: I could find no recorded rationale for `-> None` here. `Returns: None` is a blanket docstring template across ~80 catalog methods including plain voids like `delete_workspace`; there's no comment or doc note explaining it; and `-> None` dates to the original `NAS-3058` implementation rather than a later removal. The closest relative, `set_hll_type`, documents its idempotency but is silent on the outcome. That's absence of evidence, not evidence of absence — the deliberate-choice reading above is still entirely possible, and you'd know better than the history does.

## Notes on the implementation

- `UpsertOutcome(str, Enum)` rather than `enum.StrEnum` — `requires-python` is `>=3.10` and `StrEnum` landed in 3.11. Matches the existing `SortDirection(str, Enum)` precedent.
- An enum rather than a `bool`: for an "or update" operation the bool *is* the entire return value, so `if sdk.catalog_workspace.create_or_update(ws):` reads as nonsense. It also leaves room for `UNCHANGED` later.
- The three methods with an "id is `None`" path (`workspace_setting`, `user_data_filter`, `filter_view`) return `CREATED` on that branch, but see the pre-existing defect below — that branch cannot currently execute, so the docstrings describe it as unreachable rather than promising a creation.

## Testing

- Assertions on **both** branches added to the existing cassette tests (workspace, workspace setting, user data filter, jwk, user, user group, data source). **No cassette needed re-recording** — the outcome is derived from the already-recorded interactions, so the assertions double as a check that `CREATED`/`UPDATED` matches what the server actually did.
- New `tests/catalog/test_upsert_outcome.py` adds mock-based unit tests for what no cassette reaches: filter views, export templates, and the update branch of user data filters. These stub only the branch-deciding getter, since the branch is chosen by whether that GET raises `NotFoundException`.
- `ruff` 0.15.1 (the pinned pre-commit version) check + format clean; `packages/gooddata-sdk`: **498 passed, 2 skipped, 3 xfailed**.

## A pre-existing defect this turned up

Writing those tests showed that the `id is None` create branches **cannot run at all** — in `create_or_update_filter_view`, `create_or_update_user_data_filter`, and `create_or_update_workspace_setting`.

Each serializes through a generated model that requires a `str` id, and passing `None` fails type validation before any request is made:

```
ApiTypeError: Invalid type for variable 'id'. Required value type is str and passed type was NoneType at ['id']
```

This holds for the `PostOptionalId` variants too, where "optional" means *omit the key*, not *accept None*. It predates this PR — the branches were already unreachable; this change only added a `return` to them.

I have **not** fixed it here, as it's a separate concern in the entity models / generated client. It's pinned by strict `xfail` tests so that fixing it trips them and prompts a revisit, and the docstrings no longer claim those paths create anything. Happy to split it into its own issue or PR if you'd like.

## Docs

Five `create_or_update*` pages had a hardcoded `Returns: None` and are updated (two different legacy formats in use). The remaining four methods have no docs pages.

## Summary by CodeRabbit

* **New Features**
  * Create-or-update operations now report whether an item was **created** or **updated**.
  * Added the public `UpsertOutcome` type with `CREATED` and `UPDATED` values.
  * Applies across data sources, users, user groups, workspaces, settings, filters, views, JWKs, and export templates.

* **Documentation**
  * Updated SDK documentation to describe the new return values and possible outcomes.

* **Tests**
  * Added and updated coverage verifying created and updated results across supported operations.

### DIFF
--- a/packages/gooddata-sdk/src/gooddata_sdk/catalog/types.py
+++ b/packages/gooddata-sdk/src/gooddata_sdk/catalog/types.py
@@ -18,3 +18,7 @@ class UpsertOutcome(str, Enum):
 
     CREATED = "created"
     UPDATED = "updated"
+
+    # Match StrEnum's str() (the value, not "UpsertOutcome.CREATED") so moving
+    # to StrEnum once py3.10 support is dropped is a no-op for callers.
+    __str__ = str.__str__
--- a/packages/gooddata-sdk/src/gooddata_sdk/catalog/types.py
+++ b/packages/gooddata-sdk/src/gooddata_sdk/catalog/types.py
@@ -1,5 +1,20 @@
 # (C) 2022 GoodData Corporation
 from __future__ import annotations
 
+from enum import Enum
+
 # Use typing collection types to support python < py3.9
 ValidObjects = dict[str, set[str]]
+
+
+class UpsertOutcome(str, Enum):
+    """Which branch a ``create_or_update*`` method took.
+
+    The outcome is best-effort: it reports the branch the SDK chose after its
+    existence check, and that check is not atomic with the write that follows.
+    A concurrent actor can create or delete the entity in between, so treat the
+    value as informational rather than as an authoritative audit record.
+    """
+
+    CREATED = "created"
+    UPDATED = "updated"
--- a/docs/content/en/latest/administration/organization/create_or_update_jwk.md
+++ b/docs/content/en/latest/administration/organization/create_or_update_jwk.md
@@ -6,7 +6,7 @@ weight: 100
 api_ref: "CatalogOrganizationService.create_or_update_jwk"
 ---
 
-``create_or_update_jwk( jwk: CatalogJwk ) -> None``
+``create_or_update_jwk( jwk: CatalogJwk ) -> UpsertOutcome``
 
 Create a new jwk or overwrite an existing jwk with the same id.
 
@@ -18,4 +18,6 @@ Create a new jwk or overwrite an existing jwk with the same id.
 
 ## Returns
 
-_None_
+| type | description |
+| -- | -- |
+| UpsertOutcome | CREATED if the jwk did not exist yet, UPDATED if it did. |
--- a/docs/content/en/latest/administration/user-groups/create_or_update_user_group.md
+++ b/docs/content/en/latest/administration/user-groups/create_or_update_user_group.md
@@ -19,7 +19,10 @@ User group entity object.
 {{< /parameter >}}
 {{% /parameters-block %}}
 
-{{% parameters-block title="Returns" None="yes" %}}
+{{% parameters-block title="Returns" %}}
+{{< parameter p_type="UpsertOutcome" >}}
+CREATED if the user group did not exist yet, UPDATED if it did.
+{{< /parameter >}}
 {{% /parameters-block %}}
 
 ## Example
--- a/docs/content/en/latest/administration/users/create_or_update_user.md
+++ b/docs/content/en/latest/administration/users/create_or_update_user.md
@@ -19,7 +19,10 @@ User entity object.
 {{< /parameter >}}
 {{% /parameters-block %}}
 
-{{% parameters-block title="Returns" None="yes" %}}
+{{% parameters-block title="Returns" %}}
+{{< parameter p_type="UpsertOutcome" >}}
+CREATED if the user did not exist yet, UPDATED if it did.
+{{< /parameter >}}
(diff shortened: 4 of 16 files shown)


packages/gooddata-sdk/src/gooddata_sdk/catalog/types.py
```diff
@@ -1,5 +1,20 @@
 # (C) 2022 GoodData Corporation
 from __future__ import annotations
 
+from enum import Enum
+
 # Use typing collection types to support python < py3.9
 ValidObjects = dict[str, set[str]]
+
+
+class UpsertOutcome(str, Enum):
```
hkad98, line 10: Nit: we would like to eventually switch this to StrEnum – we cannot right now because we have a support for Python 3.10 which does not support StrEnum.

Consider adding the following:

```
__str__ = str.__str__
```

It should make switch to StrEnum a noop.

vondravl: Done in c618ee29 — added `__str__ = str.__str__` on `UpsertOutcome`.

Verified it makes the eventual StrEnum swap a no-op: `str(UpsertOutcome.CREATED)` / `f"{...}"` / `"%s" %` / `json.dumps` all yield `"created"` instead of `"UpsertOutcome.CREATED"`, while `repr()`, `== "created"` and `UpsertOutcome("created")` are unchanged. Pinned it in `test_upsert_outcome.py::test_outcome_is_a_plain_string` (`str(outcome) == outcome.value` + f-string), so a future base-class change that regresses this trips the test.

hkad98 (approved):

PR #1706, opened 2026-08-04
feat(gooddata-eval): add KDA-skill agentic evaluator

FrankHuynh:
## What

Adds `kda_skill.py` to `gooddata-eval`, evaluating the chatbot's `create_key_driver_analysis` / `execute_key_driver_analysis` tool calls against the `kda_skill` Langfuse dataset.

Related: [QA-28800](https://gooddata.atlassian.net/browse/QA-28800) — Build E2E LLM test for KDA skill.

## Scope

Scope is strictly completion, not field correctness:

```
strict_pass = triggered AND executed AND success AND turn_completed
```

Per-field checks (`Measure` / `Date Attribute` / `Analyzed Period` / `Reference Period` / `Filters` / `Summary` within tolerance) are **out of scope for this PR** — deferred to a follow-up ticket ([QA-28699](https://gooddata.atlassian.net/browse/QA-28699)).

## Latency / performance reporting

Latency is measured directly by the harness, not derived from a Langfuse trace after the fact. `ChatClient` times each `send_message()` call around the SSE stream (starting before the request opens, so it includes connection setup time; excludes retry backoff). `create` and `execute` are always called together within the same turn — confirmed against the skill's own system prompt on the `gdc-nas` side ("Call `create_key_driver_analysis` and then `execute_key_driver_analysis` immediately - NO confirmation needed") — so `KdaRunResult.turn_wall_clock_sec` is simply that one turn's own time (`None` if `create` never happened), excluding any earlier disambiguation turn and the simulated-reply's own OpenAI call.

`evaluate_agentic_kda_skill` logs it as the `kda_turn_wall_clock_sec` Langfuse score; `gdc-nas`'s `combo_report.py` reads it directly to classify pass/failed/error — no trace re-resolution needed.

The latency logged is the whole conversational turn (LLM planning + every tool call), not isolated KDA-backend execution time — gen-ai doesn't trace tool execution as its own observation, so there's no narrower point to measure from on either side of this cross-repo pair.

## Disambiguation safety net

KDA cases are designed to resolve in one turn. The only thing that extends a run, bounded to `max_iterations` (default 3): the agent asks a clarifying question instead of triggering KDA directly (metric or analyzed-period ambiguity) — a simulated-user reply (gpt-4o-mini, 30s timeout) picks any acceptable candidate and continues. 3, not 2 (or 1): metric and period can each need their own clarifying question, so a case ambiguous on both can legitimately take 2 rounds before `create` is ever called; still asking after that is treated as the model failing to resolve the object, not a KDA-specific failure.

`create` succeeding without `execute_result` in that same turn is scored as triggered-but-not-executed immediately, not treated as "wait for a later turn" — confirmed against the real implementation there's no code path where the agent calls `create`, stops, and needs a follow-up nudge to call `execute`. The one real case where `execute` doesn't run is the org having data-sharing with the LLM turned off, in which case `execute_key_driver_analysis` isn't registered as a tool at all for that conversation — a permanent capability gap, not a delayed call, and no chat-level nudge can work around it.

## Also fixed in review

- `pass_at_k`/`pass_power_k` Langfuse scores are `kda_`-prefixed — unprefixed, `pass_at_2` at `k=2` collides with `visualization.py`'s own score name, which `gdc-nas`'s `combo_report.py.verdict()` checks first.
- A `ChatError`/`TransientChatError` raised mid-turn no longer propagates out of `run_agentic_kda_skill` uncaught — it now surfaces as a normal failed run, so it still gets scored to Langfuse.
- `ChatError`/`TransientChatError` now carry `partial_result`, so tool calls that already succeeded before a later, unrelated error (e.g. the final summary failing) aren't discarded and misreported as "KDA never triggered".
- `turn_completed` resets to `False` in the exception branch, so a crash on a later iteration can't leave a stale `True` from an earlier one.
- `run_agentic_kda_skill` rejects `k < 1` — a bad env-driven `KDA_RUN_K` value previously ran silently once instead of surfacing the bad config.
- `stream_ended` is now set the moment the `response_ended` event line is parsed, not its data line — an event with no data payload previously left the flag unset.
- `t0` now begins before opening the SSE stream, not after — it was missing the connection/server-setup time a caller actually waits through.
- Disambiguated-turn latency no longer double-counts every turn's time plus the simulated-reply's own OpenAI call.
- The 6 per-field informational correctness scores and their support code are removed — team confirmed this PR's scope is trigger+complete only.
- `KdaEvaluation.kda_triggered` renamed to `triggered`, matching the other three core fields — the Langfuse score name `kda_triggered` is unchanged.
- `generate_simulated_kda_response`'s OpenAI call now has a 30s timeout.
- `kda_skill` wired into the CLI's agentic dispatcher (`agentic_runner.py`) and `AGENTIC_TEST_KINDS`, matching every other agentic skill, so it can be run/debugged standalone via `gd-eval run` instead of only through `gdc-nas`'s tavern-e2e harness. (README's `test_kind` table isn't updated for this — it documents a different, older evaluator subsystem and has zero `agentic_*` entries for any of the 6 existing agentic skills, so adding only `kda_skill` there would be misleading.)
- `dataset_name` default renamed from `agent_kda_skill` to `kda_skill`, matching the no-prefix pattern every other skill uses; no functional change since `gdc-nas` always passes it explicitly.
- `expected_output.get("Measure")` now guards for non-dict shapes — `DatasetItem.expected_output` on the `gdc-nas` side allows `str`/`list`, not just `dict`, and a list-shaped item previously raised `AttributeError`, silently swallowed and disabling disambiguation with only a WARNING.
- `_DEFAULT_MAX_ITERATIONS` raised from 2 to 3 — metric and period ambiguity can each need their own clarifying question, so a case ambiguous on both can legitimately take 2 rounds before `create` is ever called.
- `kda_pass_at_k`/`kda_pass_power_k` no longer logged to Langfuse — matches `metric_skill`/`alert_skill`/`guardrail`/`search_tool`/`general_question`, which all compute the pair but never log it at their default `k=1`; nothing reads a `kda_pass_at_1` score today, and the score name would shift if `k` ever changes.
- `KdaRunResult.eval` renamed to `evaluation` — it shadowed the `eval` builtin and is part of the published surface (exported in `core/agentic/__init__.py`'s `__all__`); fixed now, before `gdc-nas` starts consuming this module.
- A few comments/docstrings that had drifted from actual behavior corrected: `KdaRunResult`'s docstring, the `turn_wall_clock_sec` field comment, and `_extract_kda_calls`' docstring.

## Known limitations (out of scope for this PR)

- `cost_usd` on the KDA score is still read from `pt.total_cost`, where `pt` is the Langfuse trace picked by the default max-latency selector — not necessarily the KDA turn itself. Latency was fixed to not depend on this trace selection; cost wasn't, since this ticket's scope is latency/completion only. Candidate for a follow-up if cost accuracy on this score matters.

## Verification

No local Tiger instance available, so verified two ways:

- **Offline unit checks** against synthetic and real captured data (`ruff check`, `ruff format --check` all clean; `_evaluate_run` exercised directly with real SSE payloads captured from a 30-run manual stability test).
- Package-level import (`from gooddata_eval.core.agentic import evaluate_agentic_kda_skill, ...`) verified to resolve with no circular-import issues.

## Not included in this PR

- Version bump / release — will follow in a separate, explicitly-confirmed step (the version is monorepo-shared across all 9 published packages).
- The `gdc-nas` side (shim, tavern test, fixtures, cron wiring, daily report's KDA table) — companion PR in `gdc-nas` under the same QA-28800 ticket.

[QA-28800]: https://gooddata.atlassian.net/browse/QA-28800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QA-28699]: https://gooddata.atlassian.net/browse/QA-28699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
@@ -1,0 +1,382 @@
+# (C) 2026 GoodData Corporation. All rights reserved.
+"""Agentic KDA (Key Driver Analysis)-skill evaluation runner."""
+
+from __future__ import annotations
+
+import logging
+import os
+import re
+from dataclasses import dataclass
+
+from gooddata_eval.core.chat.sse_client import ChatClient
+from gooddata_eval.core.config import ReasoningEffort
+from gooddata_eval.core.models import ToolCallEvent
+
+_log = logging.getLogger(__name__)
+
+_DEFAULT_K = 1
+# Disambiguation safety net only (create+execute always run together in the same
+# turn) -- 3 covers metric and period each needing their own clarifying question.
+_DEFAULT_MAX_ITERATIONS = 3
+
+
+def _is_asking_kda_clarification(text: str) -> bool:
+    """True if ``text`` reads as the agent asking for input, not a final answer.
+
+    KDA-specific, not shared with metric_skill.py/conversation.py -- each skill's
+    disambiguation heuristic has already drifted independently. Requires the text to
+    end on "?" (a "?" anywhere also matches a final answer that merely quotes one).
+    """
+    if not text:
+        return False
+    t = text.strip().lower()
+    if t.endswith("?"):
+        return True
+    # "To clarify, ..." means "in other words" (a final answer), not a request for one --
+    # strip it first so "clarif" below only matches genuine clarification requests.
+    t = re.sub(r"^(just )?to clarify,?\s*", "", t)
+    return "could you" in t or "please provide" in t or "clarif" in t
+
+
+def generate_simulated_kda_response(agent_message: str, measure_candidates: dict | list[dict] | None) -> str:
+    """Generate a user reply to keep the KDA-skill conversation going (gpt-4o-mini).
+
+    Used only when the agent asks a clarifying question instead of triggering KDA
+    directly. Picks *any* candidate from ``measure_candidates`` -- scope only needs KDA
+    to trigger, not the resulting measure to be exactly right. Always OpenAI regardless
+    of the combo's own provider -- this is test-harness plumbing, not the system under test.
+    """
+    try:
+        from openai import OpenAI  # noqa: PLC0415
+    except ImportError as exc:
+        raise RuntimeError("openai package is required for generate_simulated_kda_response") from exc
+
+    api_key = os.environ.get("OPENAI_API_KEY")
+    if not api_key:
+        raise OSError("OPENAI_API_KEY environment variable is not set")
+
+    client = OpenAI(api_key=api_key)
+    candidates = measure_candidates if isinstance(measure_candidates, list) else [measure_candidates or {}]
+    candidate_desc = "; or ".join(
+        f"{c.get('type')} '{c.get('id')}'" + (f" (aggregation {c['aggregation']})" if c.get("aggregation") else "")
+        for c in candidates
+    )
+    prompt = (
+        f"You are simulating a user in a conversation with a BI assistant that runs key driver "
+        f"analysis. The assistant said: '{agent_message}'. "
+        f"The user is happy to proceed with any of the following: {candidate_desc}. "
+        f"Reply briefly as the user, picking whichever of those the assistant offered."
+    )
+    response = client.chat.completions.create(
+        model="gpt-4o-mini",
+        messages=[{"role": "user", "content": prompt}],
+        max_tokens=150,
+        timeout=30,
+    )
+    return response.choices[0].message.content or "Please proceed with either option."
+
+
+def _extract_kda_calls(tool_call_events: list[ToolCallEvent]) -> tuple[dict | None, dict | None]:
+    """Return (create_args, execute_result) for the LAST create/execute pair in this turn's
+    tool calls -- not the last create and last execute picked independently. A new create
+    call clears any earlier execute_result -- it belongs to the create it followed, not to
+    this one.
+    """
+    create_args: dict | None = None
+    execute_result: dict | None = None
+    for tc in tool_call_events:
+        if tc.function_name == "create_key_driver_analysis":
+            create_args = tc.parsed_arguments()
+            execute_result = None
+        elif tc.function_name == "execute_key_driver_analysis" and tc.result:
+            execute_result = tc.parsed_result()
+    return create_args, execute_result
+
+
+@dataclass
+class KdaEvaluation:
+    """Evaluation scores for a single KDA-skill run.
+
+    Scope: asserts only that the KDA process runs to completion -- the tool chain
+    triggers, executes successfully, and the chat turn ends cleanly with a non-empty
+    response (``turn_completed`` requires both gen-ai's stream-ended signal and a
+    non-empty ``text_response`` -- a stream that ends cleanly but delivers nothing to the
+    user isn't a completed turn either).
+    """
+
+    triggered: bool
+    executed: bool
+    success: bool
+    turn_completed: bool
+    disambiguated: bool = False
+
+    @property
+    def strict_pass(self) -> bool:
+        return all([self.triggered, self.executed, self.success, self.turn_completed])
+
+
+@dataclass
+class KdaRunResult:
+    """Outcome of one run (one conversation, up to max_iterations messages) for a KDA case."""
+
+    conversation_id: str
+    evaluation: KdaEvaluation
+    actual_create_args: dict | None
+    actual_execute_result: dict | None
+    # Wall-clock time of the turn that called create (None if create never happened) --
+    # not any earlier disambiguation turn. See run_agentic_kda_skill's _run_once.
+    turn_wall_clock_sec: float | None = None
+
+
+@dataclass
+class AgenticKdaSummary:
+    """Aggregated outcome of K runs for a KDA case."""
+
+    run_results: list[KdaRunResult]
+    pass_at_k: bool
+    pass_power_k: bool
+    best: KdaRunResult
+
+
+def _evaluate_run(
+    create_args: dict | None,
+    execute_result: dict | None,
+    turn_completed: bool,
+    disambiguated: bool = False,
+) -> KdaEvaluation:
+    triggered = create_args is not None
+    executed = execute_result is not None
+    success = executed and execute_result.get("success") is True
+    return KdaEvaluation(
+        triggered=triggered,
+        executed=executed,
+        success=success,
+        turn_completed=turn_completed,
+        disambiguated=disambiguated,
+    )
+
+
+def run_agentic_kda_skill(
+    host: str,
+    token: str,
+    workspace_id: str,
+    question: str,
+    expected_output: dict,
+    k: int = _DEFAULT_K,
+    max_iterations: int = _DEFAULT_MAX_ITERATIONS,
+    initial_conversation_id: str | None = None,
+    reasoning_effort: ReasoningEffort | None = None,
+) -> AgenticKdaSummary:
+    """Run the KDA-skill agentic evaluation K times and return a summary.
+
+    Each run is normally one message, one turn -- create and execute are always called
+    together in the same turn (the skill's own system prompt: "NO confirmation needed").
+    The only thing that can extend a run up to ``max_iterations`` turns is the agent
+    asking a clarifying question instead of triggering KDA directly; a simulated user
+    reply nudges it forward.
+    """
+    if k < 1:
+        # k=0 or negative would otherwise silently run once, indistinguishable from k=1.
+        raise ValueError(f"k must be >= 1, got {k}")
+    run_results: list[KdaRunResult] = []
+    client = ChatClient(host=host, token=token, workspace_id=workspace_id, reasoning_effort=reasoning_effort)
+
+    def _run_once(conv_id: str) -> KdaRunResult:
+        create_args: dict | None = None
+        execute_result: dict | None = None
+        turn_wall_clock_sec: float | None = None
+        turn_completed = False
+        disambiguated = False
+        current_question = question
+
+        for iteration in range(max_iterations):
+            try:
+                chat_result = client.send_message(conv_id, current_question)
+            except Exception as exc:  # noqa: BLE001 -- end this run, not the whole assertion
+                _log.warning("KDA send_message failed for conversation %s: %s", conv_id, exc)
+                partial = getattr(exc, "partial_result", None)
+                if partial is not None:
+                    create_args, execute_result = _extract_kda_calls(partial.tool_call_events or [])
+                    if create_args is not None:
+                        turn_wall_clock_sec = partial.turn_wall_clock_sec
+                turn_completed = False
+                break
+            create_args, execute_result = _extract_kda_calls(chat_result.tool_call_events or [])
+            response_text = (chat_result.text_response or "").strip()
+            turn_completed = chat_result.stream_ended and bool(response_text)
+            if create_args is not None:
+                # This turn's own time -- the turn that called create, not any earlier
+                # disambiguation turn or the simulated-reply generation. create and execute
+                # are always called together in the same turn (or not at all), so this is
+                # final either way -- execute_result may still be None (e.g. the skill's
+                # execute tool isn't available at all when data-sharing is off for the org).
+                turn_wall_clock_sec = chat_result.turn_wall_clock_sec
+                break
+            if iteration >= max_iterations - 1:
+                break
+            if _is_asking_kda_clarification(response_text):
+                measure_candidates = expected_output.get("Measure") if isinstance(expected_output, dict) else None
+                try:
+                    current_question = generate_simulated_kda_response(response_text, measure_candidates)
+                    disambiguated = True
+                except Exception as exc:  # noqa: BLE001 -- safety net, not the assertion; end only this run
+                    _log.warning("Simulated KDA user reply failed for conversation %s: %s", conv_id, exc)
+                    break
+            else:
+                break
+
+        ev = _evaluate_run(create_args, execute_result, turn_completed, disambiguated)
+        return KdaRunResult(
+            conversation_id=conv_id,
+            evaluation=ev,
+            actual_create_args=create_args,
+            actual_execute_result=execute_result,
+            turn_wall_clock_sec=turn_wall_clock_sec,
+        )
+
+    try:
+        conv_id_0 = initial_conversation_id if initial_conversation_id is not None else client.create_conversation()
+        try:
+            run_results.append(_run_once(conv_id_0))
+        finally:
+            if initial_conversation_id is None:  # only delete conversations we created
+                client.delete_conversation(conv_id_0)
+
+        for _ in range(1, k):
+            conv_id = client.create_conversation()
+            try:
+                run_results.append(_run_once(conv_id))
+            finally:
+                client.delete_conversation(conv_id)
+    finally:
+        client.close()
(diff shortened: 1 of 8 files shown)


packages/gooddata-eval/src/gooddata_eval/core/agentic/_langfuse.py
```diff
@@ -54,6 +54,30 @@ def _ts(v: Any) -> str:
...
+
+
+class _ObservationsAPI:
+    def __init__(self, client: httpx.Client) -> None:
+        self._client = client
+
+    def list(self, trace_id: str, limit: int = 100) -> _ObservationListResult:
+        resp = self._client.get("/api/public/observations", params={"traceId": trace_id, "limit": limit})
+        resp.raise_for_status()
+        return _ObservationListResult([_ObservationObj(o) for o in resp.json().get("data", [])])
```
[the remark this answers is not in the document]
FrankHuynh: Fixed — `_ObservationsAPI.list()` now accepts `page`, and `_trace_has_kda_call` pages through all observations (bounded by `_OBSERVATIONS_PAGE_CAP=10`, i.e. up to 1000) instead of stopping at the first 100, early-exiting as soon as a KDA tool call is found. Verified with a mock forcing the call onto page 2 (confirms it's actually reached, not just accepted as a param) and against the real captured traces from earlier debugging. Scoped to the legacy `/api/public/observations` endpoint (page/limit, matches this file's existing pagination style) rather than the v2 cursor API — this internal tool only targets Langfuse Cloud, so the v3/v4 self-hosted compatibility question doesn't apply here.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,423 @@
...
+            chat_result = client.send_message(conv_id, current_question)
+            c_args, e_result = _extract_kda_calls(chat_result.tool_call_events or [])
+            turn_completed = bool((chat_result.text_response or "").strip())
+            if c_args is not None:
+                create_args, execute_result = c_args, e_result
+                break
+            response_text = (chat_result.text_response or "").strip()
+            if iteration >= max_iterations - 1 or not _is_asking_clarification(response_text):
+                break
+            current_question = generate_simulated_kda_response(response_text, expected_output.get("Measure"))
```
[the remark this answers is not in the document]
FrankHuynh: Fixed exactly as proposed — wrapped the `generate_simulated_kda_response` call in try/except, logging via the module-level `_log` (already existed) and `break`-ing to end only the current run. Verified with a mock: a helper failure on run 2 of a k=2 evaluation no longer raises past `_run_once` — run 1 stays intact in `run_results`, and the failing run just resolves to `kda_triggered=False` instead of crashing the whole evaluation.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,540 @@
...
+        )
+        suffix_needed = len(summary.run_results) > 1
+        for run_idx, run in enumerate(summary.run_results):
+            pt = traces_by_conv.get(run.conversation_id)
+            run_name = f"{run_name_base}_run{run_idx}" if suffix_needed else run_name_base
+            ev = run.eval
+            # Gates strict_pass -- current scope is completion only (see KdaEvaluation docstring).
+            strict_checks = {
+                "kda_triggered": ev.kda_triggered,
+                "executed": ev.executed,
```
myhoai: Missing prefix: 
- kda_executed
- kda_success
- kda_turn_completed

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,540 @@
...
+    token: str,
+    workspace_id: str,
+    question: str,
+    expected_output: dict,
+    k: int = _DEFAULT_K,
+    max_iterations: int = _DEFAULT_MAX_ITERATIONS,
+    initial_conversation_id: str | None = None,
+    langfuse: object | None = None,
+    dataset_item_id: str = "",
+    dataset_name: str = "kda_skill",
```
myhoai, line 287: Is kda_skill the correct dataset name?

FrankHuynh: Fixed — default changed to `agent_kda_skill`, matching the actual Langfuse dataset name used everywhere else (fixtures, gdc-nas's report scripts).

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
FrankHuynh: Fixed — renamed to `kda_executed`/`kda_success`/`kda_turn_completed`, matching `kda_triggered`'s existing prefix convention. Updated `gdc-nas`'s `combo_report.py` `_KDA_STRICT_FIELDS` in the same change, since it reads these score names back for the Slack report's pass/fail determination.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,610 @@
...
+    unknown = sorted(k for k in expected_summary if "tolerance" in k.lower() and k not in _SUMMARY_TOLERANCE_KEYS)
+    if unknown:
+        _log.warning(
+            "KDA dataset item's Summary has unrecognized tolerance key(s) %s -- ignored, "
+            "falling back to the default relative tolerance. Recognized keys: %s",
+            unknown,
+            sorted(_SUMMARY_TOLERANCE_KEYS),
+        )
+    if "absolute_tolerance" in expected_summary:
+        return "absolute", expected_summary["absolute_tolerance"]
```
myhoai: `abs_tolerance` is passed through raw while `actual`/`expected` go through `_to_number()` — a string value like `"1E-2"` would raise `TypeError` in `_within_absolute_tolerance`; suggest `_to_number()` on both returns.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,610 @@
...
+                        turn_wall_clock_sec = elapsed_sec + (partial.turn_wall_clock_sec or 0.0)
+                turn_completed = False
+                break
+            elapsed_sec += chat_result.turn_wall_clock_sec or 0.0
+            c_args, e_result = _extract_kda_calls(chat_result.tool_call_events or [])
+            response_text = (chat_result.text_response or "").strip()
+            turn_completed = chat_result.stream_ended and bool(response_text)
+            if c_args is not None:
+                create_args, execute_result = c_args, e_result
+                turn_wall_clock_sec = elapsed_sec
```
myhoai: `elapsed_sec` here is turn 1 + the `generate_simulated_kda_response` OpenAI call (l.431) + turn 2, but `_KDA_MAX_TURN_SEC` in `combo_report.py` is a single-turn threshold — a disambiguated case whose KDA turn took 45s can be reported as a 67s breach, and the harness's own gpt-4o-mini latency (no `timeout=` set) counts as product latency.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,610 @@
...
+    Attribute/Periods/Filters/Summary matching the expected values) is computed and
+    logged for visibility but intentionally excluded from ``strict_pass`` -- that
+    verification is scoped to a follow-up ticket, not this one.
+    """
+
+    # Core: gates strict_pass.
+    kda_triggered: bool
+    executed: bool
+    success: bool
+    turn_completed: bool
```
myhoai: `kda_triggered` carries the prefix but the other three don't — suggest `triggered` for consistency, since the Langfuse score names are uniformly `kda_*` anyway.

packages/gooddata-eval/tests/test_sse_client.py
```diff
@@ -205,6 +320,40 @@ def handler(request):
...
 
+def test_send_message_sets_turn_wall_clock_sec_on_success(monkeypatch):
+    monkeypatch.setattr(sse_mod.time, "monotonic", iter([100.0, 102.5]).__next__)
+    client = _client_with_handler(lambda request: httpx.Response(200, content=_OK_SSE))
+    result = client.send_message("conv", "q")
+    assert result.turn_wall_clock_sec == pytest.approx(2.5)
+
+
+def test_send_message_wall_clock_excludes_retry_backoff(monkeypatch):
+    # t0 must be per-attempt, set inside _do() after the connection is already open --
```
myhoai: "after the connection is already open" is stale — `t0` is set before `self._client.stream(...)`, so connection setup is included on purpose.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,386 @@
...
+class AgenticKdaSummary:
+    """Aggregated outcome of K runs for a KDA case."""
+
+    run_results: list[KdaRunResult]
+    pass_at_k: bool
+    pass_power_k: bool
+    best: KdaRunResult
+
+
+def _evaluate_run(
```
myhoai, line 141: `KdaRunResult` is exported in `core/agentic/__init__.py`'s `__all__`, so this field name is part of the published surface. It also shadows the `eval` builtin. Renaming it after this ships is a breaking change for `gdc-nas`, which is about to start consuming this module.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,386 @@
...
+    if not summary.pass_at_k:
+        best = summary.best
+        ev = best.eval
+        message = (
+            f"KDA skill assertion failed. strict_pass={ev.strict_pass} "
+            f"(triggered={ev.triggered}, executed={ev.executed}, "
+            f"success={ev.success}, turn_completed={ev.turn_completed}). "
+            f"Actual create args: {best.actual_create_args}. "
+            f"Actual execute result: {best.actual_execute_result}."
+        )
```
myhoai, line 381: At `k=1` both values are identical to `strict_pass`, and nothing reads them: `combo_report.py` does not reference `kda_pass_at_*` anywhere, and no workflow sets `KDA_RUN_K`. Separately, the score *name* varies with `k`, so raising `KDA_RUN_K` later splits the data into two series that cannot be joined, and silently breaks any Langfuse view built on the previous name.

Could you double-check this against how the other skills handle the same pair? From what I can see, `metric_skill`, `alert_skill`, `general_question`, `guardrail` and `search_tool` all compute `pass_at_k`/`pass_power_k` but never log them — all of them have `_DEFAULT_K = 1`. `visualization.py` is the only one that logs them (`:335-336`), and it runs `_DEFAULT_K = 2` with a real consumer (`combo_report`'s `viz_flaky` reads the `pass_at_2` / `pass_power_2` pair). It also derives the name from `len(summary.run_results)` (`:313`) rather than the requested `k`. I'd like to understand whether KDA is intentionally diverging from both of those, or whether this followed `visualization` without the `K=2` that justifies it.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,386 @@
...
+        conv_id_0 = initial_conversation_id if initial_conversation_id is not None else client.create_conversation()
+        try:
+            run_results.append(_run_once(conv_id_0))
+        finally:
+            if initial_conversation_id is None:  # only delete conversations we created
+                client.delete_conversation(conv_id_0)
+
+        for _ in range(1, k):
+            conv_id = client.create_conversation()
+            try:
```
myhoai, line 247: `expected_output.get("Measure")` assumes a dict, but `DatasetItem.expected_output` on the `gdc-nas` side allows `str | list[str] | dict | list[dict]`. A list-shaped item raises `AttributeError` here, which is swallowed by the `except Exception` below and leaves only a WARNING — disambiguation is disabled for that case with the suite still green.

The same silence applies if the Langfuse key stops being exactly `"Measure"`: `candidate_desc` degrades to `"None 'None'"` and the simulated-user prompt becomes meaningless without any signal.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+@dataclass
+class AgenticKdaSummary:
+    """Aggregated outcome of K runs for a KDA case."""
+
+    run_results: list[KdaRunResult]
+    pass_at_k: bool
+    pass_power_k: bool
+    best: KdaRunResult
```
myhoai, line 138: "one conversation, one message" no longer matches the code — a run is now up to `max_iterations` messages (disambiguation reply, and the continuation nudge at `:242`). The multi-turn behaviour this docstring contradicts is introduced by this PR.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+def _evaluate_run(
+    create_args: dict | None,
+    execute_result: dict | None,
+    turn_completed: bool,
+    disambiguated: bool = False,
+) -> KdaEvaluation:
```
myhoai, line 146: The comment says "the specific `send_message()` call that **triggered** KDA", but the assignment at `:232-235` happens only once **both** `create` and `execute` have resolved — i.e. the turn that *completed* KDA, not the one that triggered it.

This PR is otherwise careful to keep `triggered` and `executed` distinct, and this is the one place the distinction matters most given the agreed latency definition. `tests/test_agentic_kda_skill.py:430` (`test_run_agentic_kda_skill_latency_is_only_the_turn_that_triggered_kda`) carries the same wording.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -0,0 +1,386 @@
...
+    return create_args, execute_result
+
+
+@dataclass
+class KdaEvaluation:
+    """Evaluation scores for a single KDA-skill run.
+
+    Scope: asserts only that the KDA process runs to completion -- the tool chain
+    triggers, executes successfully, and the chat turn ends cleanly with a non-empty
+    response (``turn_completed`` requires both gen-ai's stream-ended signal and a
```
myhoai, line 102: This describes returning "the LAST create/execute *pair*" and explains the reset that prevents mispairing. That holds within a single turn, but the caller merges results across turns (`:228-229`), so the pairing guarantee does not extend to the value the caller ends up with.

(I traced this — it does not produce an actual mispairing, because the loop breaks as soon as both are present. Documentation only.)

myhoai (commented):
Review notes — items I'd like resolved before merge. All are scoped to this PR; nothing here disputes the agreed scope (completion-only, whole-turn E2E latency). Comments inline.

myhoai (approved):

PR #1713, opened 2026-08-05
feat(gooddata-eval): capture conversation_id/response_id through the agentic-CLI path

Tomkess:
## Summary
- Stacked on #1708 — targets `feat/chat-client-reasoning-steps`, not master.
- Same gap #1708 fixed for `reasoning_steps`, applied to `conversation_id`/`response_id`: `cli/agentic_runner.py -> evaluate_agentic_*` never surfaced them onto `ItemReport` for the three agentic kinds (`agentic_alert_skill`/`agentic_metric_skill`/`agentic_conversation`), even though each `ChatResult` already carries both.
- Widens each `evaluate_agentic_*` return from `list[str]` to `(reasoning_steps, conversation_id, response_id)`; attaches all three to the raised exception on failure (mirrors the existing `conversation_id`-on-exception idiom in `ChatClient.ask()`); `run_agentic_items` unpacks either form.
- `response_id` is new at the `RunResult` layer for all three kinds — captured as the last non-null value across a run's turns, same accumulation pattern already used for `reasoning_steps`.
- `general_question`/`guardrail`/`search_tool`/`visualization` untouched — already populated via the single-turn `runner.py` path, unaffected by this gap.

## Test plan
- [x] `uv run pytest tests/test_agentic_alert_skill.py tests/test_agentic_conversation.py tests/test_agentic_metric_skill.py tests/test_agentic_runner.py tests/test_alert_skill_evaluator.py tests/test_metric_skill_evaluator.py` — 65 passed
- [x] Full package suite (`uv run pytest`) — same 9 pre-existing failures with or without this patch (missing `openai` in local env), zero new failures
- [x] `ruff check` clean on all touched files
- [x] `ty check` — same 4 pre-existing `openai`-unresolved-import diagnostics with or without this patch, zero new diagnostics

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
+++ b/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
@@ -83,11 +83,12 @@ def _dispatch_agentic(
     run_ts: str,
     model_version_override: str | None,
     reasoning_effort: ReasoningEffort | None = None,
-) -> list[str] | None:
+) -> tuple[list[str], str, str | None] | list[str] | None:
     """Call the appropriate evaluate_agentic_* function for the item's test_kind.
 
-    Returns whatever that function returns -- only alert_skill/metric_skill/conversation
-    currently return their reasoning_steps; the rest still return None (unchanged).
+    Returns whatever that function returns -- alert_skill/metric_skill/conversation return
+    their (reasoning_steps, conversation_id, response_id); the rest still return None
+    (unchanged).
     """
     kind = item.test_kind
     eo = item.expected_output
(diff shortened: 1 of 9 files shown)


packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
```diff
@@ -85,11 +85,12 @@ def _dispatch_agentic(
     run_ts: str,
     model_version_override: str | None,
     reasoning_effort: ReasoningEffort | None = None,
-) -> list[str] | None:
+) -> tuple[list[str], str, str | None] | list[str] | None:
```
hkad98: Create a dedicated dataclass for this output please

Tomkess: Fixed in dcf9f5b9 — added `AgenticEvalOutcome` (reasoning_steps, conversation_id, response_id) to `core/models.py` and made `evaluate_agentic_alert_skill`/`metric_skill`/`conversation` return it instead of the bare tuple. `_dispatch_agentic`'s return type is now `AgenticEvalOutcome | list[str] | None`, and the `isinstance(outcome, tuple)` unpack in `run_agentic_items` is now `isinstance(outcome, AgenticEvalOutcome)`. Updated the tests that unpacked the old tuple.

PR #1722, opened 2026-08-07
refactor: remove support for Greenplum

jeskepetr:
## Summary

Greenplum is no longer used by any customer and is being removed from the backend ([F1-2435](https://gooddata.atlassian.net/browse/F1-2435), gdc-nas [#25474](https://github.com/gooddata/gdc-nas/pull/25474)), so the SDK no longer needs to model it.

**Hand-written SDK**
- Removed `CatalogDataSourceGreenplum` and `GreenplumAttributes` and their exports from `gooddata_sdk`.
- Removed the Greenplum example from the data-source docs.
- Removed the Greenplum test. It was already inert — the whole function sat inside a `"""` block, and the `greenplum.yaml` cassette it referenced does not exist in the repo.

**Generated client** — updated the supported way, not by hand:
- Dropped the `GREENPLUM` enum entry from the two source schemas it originates from, `gooddata-metadata-client.json` and `gooddata-scan-client.json`.
- Regenerated `gooddata-api-client/` with openapi-generator against the merged schema.

The regeneration produced **exactly six deletions and no other drift**, which incidentally confirms the checked-in client was in sync with its schemas.

## A note on `jq`

`schemas/gooddata-api-client.json` was edited line-wise rather than rebuilt through the Makefile's `cat | jq -S -s` merge step.

jq 1.6 does not preserve number literals and rewrote unrelated values across the file — `2.6E+9` → `2600000000`, `1.0` → `1`, `100.0` → `100`. The checked-in copy was clearly produced with jq **1.7+**, which does preserve them. The line-wise edit yields exactly what the merge would produce on jq 1.7, without the collateral churn.

Worth knowing for anyone regenerating: **run `make api-client` with jq 1.7+**, or you will commit spurious numeric diffs.

## Test Plan

- Pre-commit hooks pass on the change: ruff (lint), ruff format, copyright, trailing whitespace, EOF.
- All three edited Python files compile.
- Repo-wide sweep for `greenplum` (case-insensitive) returns zero hits; nothing else in the repo imported either removed class.
- Both edited source schemas and the merged schema validate as JSON.
- openapi-generator ran clean; the resulting diff is 6 deletions of a single enum line each, with no untracked or deleted files.

**Not run: the test suite.** The local toolchain could not run it — `uv` 0.10.7 against the repo's `~=0.11.0` pin, and the repo venv is missing `gooddata_code_convertors`. I confirmed that failure reproduces on a clean tree, so it predates this change, but it does mean CI is the first real execution of these tests.

## Risk

Low in effect, but note this is a **breaking change for a published package**: anyone with `from gooddata_sdk import CatalogDataSourceGreenplum` (or `GreenplumAttributes`) gets an `ImportError` on upgrade — at import time, even in code that never connects to Greenplum.

Since nobody is using the datasource type, no working code should construct these. But if this repo prefers a deprecation cycle for public symbols, the alternative is to keep both names as deprecated shims now and remove them in the next major. Happy to switch to that if reviewers prefer.

`test_allowed_data_source_type` iterates the generated `JsonApiDataSourceInAttributes.allowed_values` and constructs the generic `CatalogDataSource`, so it is unaffected by the subclass removal and simply stops seeing `GREENPLUM`.

## Ordering

This can land independently of the backend PR. Until gdc-nas ships, the deployed backend still accepts `GREENPLUM`; afterwards it will not. Nothing here depends on deploy order.

[F1-2435]: https://gooddata.atlassian.net/browse/F1-2435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Breaking Changes**
  - Removed Greenplum as a supported data source type across API clients, SDK models, and metadata schemas.
  - Greenplum-specific SDK data source classes and public exports are no longer available.
- **Documentation**
  - Removed the Greenplum data source example from the documentation.
- **Tests**
  - Removed obsolete Greenplum data source test coverage.


jeskepetr:
Superseded by #1723, which runs the same commit from a branch in this repository instead of my fork.

The `netlify-deploy` check here failed with `Refusing to check out fork pull request code from a 'pull_request_target' workflow` — an `actions/checkout` security guard that fires for any fork PR, unrelated to these changes. Reopening from an in-repo branch so the docs preview can build, which matters here since this PR edits the data-source docs.

PR #1728, opened 2026-08-11
refactor: remove support for Drill

jeskepetr:
## Summary

Removes `DRILL` from the data source type enums. Drill is being removed from the backend
(gdc-nas #25522), so the SDK no longer needs to model it. Mirrors the Greenplum removal in
#1723.

Unlike Greenplum there was never a hand-written `CatalogDataSourceDrill` class, so there is
nothing to drop from the `gooddata_sdk` exports, and no test or docs example to remove — this
change is narrower than #1723 by exactly those pieces.

- drop the `DRILL` enum entry from the two source schemas it originates from,
  `gooddata-metadata-client.json` (5) and `gooddata-scan-client.json` (1), and from the merged
  `gooddata-api-client.json` (6)
- drop it from the six generated models that mirror them — one line each, exactly the shape
  regeneration produces (the same 6/5/1 shape #1723 produced)
- reword two prose mentions of Drill as a multi-schema data source manager to Dremio only

## Test Plan

- all three schemas still parse as JSON; all six generated models still compile
- pre-commit passed (ruff, ruff format, copyright, trailing whitespace, EOF)
- `ENABLE_DRILL_TO_URL_BY_DEFAULT` is a drill-to-URL **setting**, unrelated to Apache Drill —
  left untouched, still 25 occurrences

The merged `schemas/gooddata-api-client.json` was edited line-wise rather than rebuilt through
the Makefile's `jq -S -s` step: jq 1.6 does not preserve number literals and rewrites unrelated
values (`2.6E+9` → `2600000000`, `1.0` → `1`) across the file, while the checked-in copy was
produced with jq 1.7+. Same approach as #1723.

## Risk

**low** — removes one value from generated enums plus schema definitions. No hand-written class
or export changes, so unlike #1723 there is no import-time breakage for anyone.

JIRA: https://gooddata.atlassian.net/browse/F1-2434


jeskepetr:
Superseded by #1729, which is opened from a branch in this repository rather than my fork. `netlify-deploy` cannot access its secret from a fork PR, so the pipeline could not go green here. Same commit, rebased onto current master.

PR #1733, opened 2026-08-12
fix(gooddata-eval): retry KDA on any non-triggering response, drop text classification

FrankHuynh:
## What

`_is_asking_kda_clarification` tried to classify agent responses as "asking for clarification" vs "a final answer" -- a simulated user reply was only sent when the text matched. That heuristic missed a real, common response shape: a clarifying question immediately followed by a bullet list of the options being offered, e.g.:

```
I found two different "Total Net Revenue" metrics in your data model.
Which one should I analyze for the 2024 vs 2023 drop?

- {metric/metric_l1_sql_net_sales_summary_net_revenue}
- {metric/metric_l1_total_net_revenue}
```

The message doesn't end on `"?"` itself (it ends on the list), so the harness treated this as a final answer and gave up after turn 1 -- never nudging the simulated user to pick an option. The case scored `triggered=False`, a hard failure, even though one of the two offered metrics was the expected answer.

## How found

Real CI trace (QA-28800, `gpt56luna_openai` combo, `globalmart` workspace): [trace `ba72a6d2c19a3ceeaa354ec641bc52f3`](https://us.cloud.langfuse.com/project/cm71tjc2a0007ad07qkyw5x9q/traces?filter=id%3Bstring%3B%3Bcontains%3Bba72a6d2c19a3ceeaa354ec641bc52f3). `kda_disambiguated` stayed `False` and the Langfuse session had only 1 trace total -- confirming the simulated-reply path was never reached.

## Fix (redesigned per review)

A first pass patched the heuristic directly (wider bullet-marker support, "?" no longer needing to be the literal last character, plus a `period_hint` gap and a `None 'None'` prompt bug fixed alongside). Review (chi My) pointed out the cost of the two error directions here is asymmetric: missing a genuine clarifying question hard-fails the run, while misreading a final answer as one only costs one harmless extra turn -- the loop already breaks for good once `create_args` is set, so a false positive can never turn a pass into a fail. Every other skill in this package (`visualization.py`, `alert_skill.py`) already solves this the cheap way: never classify the text at all, just break on the goal signal (tool called / artifact produced) or an empty response, and otherwise always retry. Patching the KDA-specific heuristic for yet another response shape (this round: `**Option 1**: ...`, bold markdown with no space after the marker) would have meant chasing an open-ended list of shapes forever.

**`_is_asking_kda_clarification` and `_LIST_ITEM_RE` are removed entirely.** `_run_once` now matches `visualization.py`/`alert_skill.py`'s own break conditions -- `create_args` set, or an empty response -- and otherwise always sends a simulated reply, regardless of what the agent's text says or how it's formatted. `_DEFAULT_MAX_ITERATIONS` bumped `3 -> 4` (3 was sized exactly for 2 real questions -- metric and period -- with zero slack for a wasted turn; every other skill in the package budgets 4-7).

Also fixed along the way:
- `generate_simulated_kda_response` only ever knew about measure candidates, even when the agent's question was about the **period** to compare instead -- it had nothing period-specific to answer with. Extracted into `_build_period_hint()`, built from whichever of `expected_output`'s `Date Attribute`/`Analyzed Period`/`Reference Period` fields are present (not requiring all three).
- The prompt asserted `"an acceptable metric/fact is None 'None'"` as a real option when `measure_candidates` was `None`/empty (e.g. a period-only question) -- likely to make gpt-4o-mini invent a metric literally named "None". Extracted into `_build_clarification_prompt()`, which now omits the "For reference, ..." clause entirely when there's nothing usable to reference.

## Tests

`_is_asking_kda_clarification`'s own unit tests removed along with the function. The end-to-end `run_agentic_kda_skill` regression tests for the real captured trace and the period-clarification case stay (now exercising the always-retry path instead of a classifier match), plus:
- A new test for the bold-markdown case chi My's review flagged (`**Option 1**: ...`).
- Direct unit tests for `_build_period_hint` (every field-presence combination) and `_build_clarification_prompt` (the `None`-candidates fix).

48 tests in `test_agentic_kda_skill.py`, all passing.

## Verification

- `ruff check` / `ruff format --check`: clean.
- `pytest packages/gooddata-eval/tests/`: 349 passed, same 9 pre-existing unrelated failures (missing `openai` module in this venv) as on `master`.

## Not included in this PR

- Version bump / release -- separate step, as with #1706.

JIRA: [QA-28800](https://gooddata.atlassian.net/browse/QA-28800)

[QA-28800]: https://gooddata.atlassian.net/browse/QA-28800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of clarifying questions followed by bulleted, numbered, or table-based options.
  - Preserved clarification detection when questions end directly with a question mark.
  - Prevented option lists followed by explanatory text from being misclassified.
  - Improved metric and comparison-period disambiguation in follow-up interactions.
  - Follow-up responses now better account for available metric and period context.
  - Increased resilience when additional clarification is needed during KDA analysis.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 
 import logging
 import os
-import re
 from dataclasses import dataclass
 
 from gooddata_eval.core.chat.sse_client import ChatClient
@@ -15,36 +14,79 @@ from gooddata_eval.core.models import ToolCallEvent
 _log = logging.getLogger(__name__)
 
 _DEFAULT_K = 1
-# Disambiguation safety net only (create+execute always run together in the same
-# turn) -- 3 covers metric and period each needing their own clarifying question.
-_DEFAULT_MAX_ITERATIONS = 3
+# Disambiguation safety net only (create+execute always run together in the same turn) --
+# 3 real questions' worth (metric, period, +1 slack) since a simulated reply is now sent on
+# every non-final turn (see run_agentic_kda_skill), not just ones classified as a question.
+_DEFAULT_MAX_ITERATIONS = 4
 
 
-def _is_asking_kda_clarification(text: str) -> bool:
-    """True if ``text`` reads as the agent asking for input, not a final answer.
-
-    KDA-specific, not shared with metric_skill.py/conversation.py -- each skill's
-    disambiguation heuristic has already drifted independently. Requires the text to
-    end on "?" (a "?" anywhere also matches a final answer that merely quotes one).
+def _build_period_hint(expected_output: dict) -> str | None:
+    """Build a period hint from whichever of expected_output's Date Attribute/Analyzed
+    Period/Reference Period are present -- a question about only one of them (e.g. "which
+    date dimension?") must still get an answerable hint, not None just because the other
+    two are absent.
     """
-    if not text:
-        return False
-    t = text.strip().lower()
-    if t.endswith("?"):
-        return True
-    # "To clarify, ..." means "in other words" (a final answer), not a request for one --
-    # strip it first so "clarif" below only matches genuine clarification requests.
-    t = re.sub(r"^(just )?to clarify,?\s*", "", t)
-    return "could you" in t or "please provide" in t or "clarif" in t
+    date_attr = expected_output.get("Date Attribute")
+    analyzed = expected_output.get("Analyzed Period")
+    reference_period = expected_output.get("Reference Period")
+    if not (date_attr or analyzed or reference_period):
+        return None
+    parts = []
+    if date_attr:
+        parts.append(date_attr)
+    if analyzed and reference_period:
+        parts.append(f"comparing {analyzed} to {reference_period}")
+    elif analyzed:
+        parts.append(f"period {analyzed}")
+    elif reference_period:
+        parts.append(f"compared to {reference_period}")
+    return ", ".join(parts)
 
 
-def generate_simulated_kda_response(agent_message: str, measure_candidates: dict | list[dict] | None) -> str:
+def _build_clarification_prompt(
+    agent_message: str, measure_candidates: dict | list[dict] | None, period_hint: str | None
+) -> str:
+    """Build the simulated-user prompt, referencing only whatever candidates/period-hint
+    are actually usable -- an empty/None candidate must drop the "acceptable metric/fact"
+    clause entirely rather than assert a literal "None" as if it were a real option.
+    """
+    candidates = [
+        c for c in (measure_candidates if isinstance(measure_candidates, list) else [measure_candidates]) if c
+    ]
+    reference = ""
+    if candidates:
+        candidate_desc = "; or ".join(
+            f"{c.get('type')} '{c.get('id')}'" + (f" (aggregation {c['aggregation']})" if c.get("aggregation") else "")
+            for c in candidates
+        )
+        reference = f"an acceptable metric/fact is {candidate_desc}"
+    if period_hint:
+        reference = (
+            f"{reference}; the intended time period is {period_hint}"
+            if reference
+            else f"the intended time period is {period_hint}"
+        )
+    return (
+        f"You are simulating a user in a conversation with a BI assistant that runs key driver "
+        f"analysis. The assistant asked: '{agent_message}'. "
+        + (f"For reference, {reference}. " if reference else "")
+        + "Reply briefly as the user, answering whichever of those the assistant actually asked about."
+    )
+
+
+def generate_simulated_kda_response(
+    agent_message: str,
+    measure_candidates: dict | list[dict] | None,
+    period_hint: str | None = None,
+) -> str:
     """Generate a user reply to keep the KDA-skill conversation going (gpt-4o-mini).
 
-    Used only when the agent asks a clarifying question instead of triggering KDA
-    directly. Picks *any* candidate from ``measure_candidates`` -- scope only needs KDA
-    to trigger, not the resulting measure to be exactly right. Always OpenAI regardless
-    of the combo's own provider -- this is test-harness plumbing, not the system under test.
+    Called on any turn that didn't trigger KDA, whatever the agent's response actually
+    said -- most often a clarifying question about the measure, the period, or both, so
+    both are given as reference and the reply answers whichever was actually asked.
+    Scope only needs KDA to trigger, not the resulting measure/period to be exactly
+    right. Always OpenAI regardless of the combo's own provider -- this is
+    test-harness plumbing, not the system under test.
     """
     try:
         from openai import OpenAI  # noqa: PLC0415
@@ -56,17 +98,7 @@ def generate_simulated_kda_response(agent_message: str, measure_candidates: dict
         raise OSError("OPENAI_API_KEY environment variable is not set")
 
     client = OpenAI(api_key=api_key)
-    candidates = measure_candidates if isinstance(measure_candidates, list) else [measure_candidates or {}]
-    candidate_desc = "; or ".join(
-        f"{c.get('type')} '{c.get('id')}'" + (f" (aggregation {c['aggregation']})" if c.get("aggregation") else "")
-        for c in candidates
-    )
-    prompt = (
-        f"You are simulating a user in a conversation with a BI assistant that runs key driver "
-        f"analysis. The assistant said: '{agent_message}'. "
-        f"The user is happy to proceed with any of the following: {candidate_desc}. "
-        f"Reply briefly as the user, picking whichever of those the assistant offered."
-    )
+    prompt = _build_clarification_prompt(agent_message, measure_candidates, period_hint)
     response = client.chat.completions.create(
         model="gpt-4o-mini",
         messages=[{"role": "user", "content": prompt}],
@@ -171,9 +203,12 @@ def run_agentic_kda_skill(
 
     Each run is normally one message, one turn -- create and execute are always called
     together in the same turn (the skill's own system prompt: "NO confirmation needed").
-    The only thing that can extend a run up to ``max_iterations`` turns is the agent
-    asking a clarifying question instead of triggering KDA directly; a simulated user
-    reply nudges it forward.
+    A run only extends past turn 1, up to ``max_iterations``, when the agent's response
+    has no create call and isn't empty; a simulated user reply is then always sent, with
+    no attempt to classify whether the text was actually asking for input (matching
+    visualization.py/alert_skill.py's own break conditions) -- missing a genuine
+    clarifying question hard-fails the run, while sending one after an unrecognized final
+    answer only costs one harmless extra turn, so the asymmetry favors never guessing.
     """
     if k < 1:
         # k=0 or negative would otherwise silently run once, indistinguishable from k=1.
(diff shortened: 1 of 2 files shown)


packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -20,31 +20,47 @@
 _DEFAULT_MAX_ITERATIONS = 3
 
 
+_LIST_ITEM_RE = re.compile(r"^\s*([-*]|\d+[.)])\s+\S")
```
myhoai: `_LIST_ITEM_RE` only covers `-`, `*`, `1.` and `1)`. Chatbot option lists in the wild also come as `•`, `–`, `+`, or as a markdown table — all of those fall through to `False` and reproduce QA-28800 verbatim. Also, `\s+\S` after the marker means a wrapped list item whose continuation line is indented without a marker fails the `all(...)` check and sinks the whole match.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -20,31 +20,47 @@
...
     if not text:
         return False
     t = text.strip().lower()
     if t.endswith("?"):
         return True
+    if "?" in t:
+        tail = t.rsplit("?", 1)[1]
+        lines = [ln for ln in tail.splitlines() if ln.strip()]
+        if lines and all(_LIST_ITEM_RE.match(ln) for ln in lines):
+            return True
```
myhoai: `rsplit("?", 1)[1]` requiring *every* remaining line to be a list item is stricter than the PR description's rationale needs. The single most common near-miss is a closing sentence after the list:

```
Which one should I analyze?
- {metric/a}
- {metric/b}

Let me know which one you'd prefer.
```

That trailing line is not a list item, so this returns `False`, and none of the `could you` / `please provide` / `clarif` fallbacks catch "let me know" either — the exact failure mode this PR is fixing.

The new branch also *adds* a false positive in the other direction: `"So, why did net revenue drop in 2024?\n\n- south -12%\n- product a -8%"` now matches. That one is fine (a false positive costs one turn), which is really the argument for relaxing this whole function rather than tightening it further.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -61,11 +77,14 @@ def generate_simulated_kda_response(agent_message: str, measure_candidates: dict
         f"{c.get('type')} '{c.get('id')}'" + (f" (aggregation {c['aggregation']})" if c.get("aggregation") else "")
         for c in candidates
     )
+    reference = f"an acceptable metric/fact is {candidate_desc}"
```
myhoai: Pre-existing, but this rewording makes it sharper: when `measure_candidates` is `None`, line 75 yields `candidates = [{}]` and `candidate_desc` becomes the literal string `None 'None'`. The old prompt said "the user is happy to proceed with any of the following: None 'None'"; the new one asserts "an acceptable metric/fact is None 'None'", which is more likely to make gpt-4o-mini invent a metric name. Worth dropping the clause entirely when there are no usable candidates — `test_run_agentic_kda_skill_disambiguates_when_expected_output_is_not_a_dict` exercises exactly this path.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
```diff
@@ -216,8 +235,15 @@ def _run_once(conv_id: str) -> KdaRunResult:
                 break
             if _is_asking_kda_clarification(response_text):
                 measure_candidates = expected_output.get("Measure") if isinstance(expected_output, dict) else None
+                period_hint = None
+                if isinstance(expected_output, dict):
+                    date_attr = expected_output.get("Date Attribute")
+                    analyzed = expected_output.get("Analyzed Period")
+                    reference_period = expected_output.get("Reference Period")
+                    if date_attr and analyzed and reference_period:
+                        period_hint = f"{date_attr}, comparing {analyzed} to {reference_period}"
```
myhoai: The `and` chain means the hint is built only when all three of `Date Attribute`, `Analyzed Period` and `Reference Period` are present. A dataset item carrying only `Date Attribute` (agent asks "which date dimension should I use?") gets `period_hint = None` and reproduces precisely the gap this PR is closing, just narrower. Suggest building the hint from whichever fields are present, e.g. emit the date attribute alone when the two periods are missing.

packages/gooddata-eval/tests/test_agentic_kda_skill.py
```diff
@@ -427,6 +427,90 @@ def test_run_agentic_kda_skill_marks_disambiguated_after_a_simulated_reply():
     assert summary.best.evaluation.triggered is True
 
 
+def test_run_agentic_kda_skill_disambiguates_on_question_followed_by_option_list():
```
myhoai, line 439: Both new tests are end-to-end through `run_agentic_kda_skill`, so the new branch is only covered indirectly and only in the positive direction. The claim in the PR description — *"real prose after the list still means this was a final answer"* — has no test at all. Two one-liners next to the existing `_is_asking_kda_clarification` block (around line 88) would lock in both edges:

```python
def test_is_asking_kda_clarification_true_on_question_followed_by_option_list():
    assert _is_asking_kda_clarification("Which metric?\n- {metric/a}\n- {metric/b}") is True


def test_is_asking_kda_clarification_false_on_option_list_followed_by_prose():
    text = "You asked which drivers moved?\n- south -12%\n- product a -8%\nThat is the full breakdown."
    assert _is_asking_kda_clarification(text) is False
```

myhoai (commented):
Reviewed the full `core/agentic/` package to put this fix in context. The change is correct and minimal — the new branch matches the captured trace, and none of the five existing `_is_asking_kda_clarification` tests regress. Two things I'd like to discuss before merge.

### 1. The cost of the two error directions is very asymmetric, and the heuristic is biased the wrong way

`_is_asking_kda_clarification` is only ever reached when `create_args is None` (`kda_skill.py:226` already broke out otherwise), i.e. only on runs that are currently failing.

- **False negative** (agent is asking, we think it answered) → break after turn 1 → `triggered=False` → hard failure. That's QA-28800.
- **False positive** (agent answered, we think it's asking) → one extra turn + one gpt-4o-mini call, plus noise on `kda_disambiguated`. It cannot turn a pass into a fail, because the loop already broke once `create_args` was set.

So false negatives are expensive and false positives are nearly free — but the matcher is written strict-first (`endswith("?")`), which optimises for the cheap direction.

I ran the post-fix heuristic against neighbouring response shapes:

| Response shape | Result |
|---|---|
| `Which one?\n- a\n- b` (the captured trace) | ✅ True |
| `Which one?\n- a\n- b\n\nLet me know which one you prefer.` | ❌ **False** |
| question + list where an item wraps onto an indented continuation line | ❌ **False** |
| `•` or `–` bullets instead of `-`/`*` | ❌ **False** |
| options rendered as a markdown table | ❌ **False** |
| `**Option 1** …` (bold, no space after `*`) | ❌ **False** |

Row 2 — question, option list, one closing sentence — is a very common LLM shape and is my best guess at the next ticket. The fix patches one shape; the adjacent ones still fail identically.

### 2. Every other skill in this package solves this differently — KDA is the only outlier

| Module | Break condition | "Is it asking?" detection | Budget |
|---|---|---|---|
| `general_question`, `guardrail`, `search_tool` | — | none, single turn | 1 |
| `visualization.py:165-179` | viz produced, **or** empty `text_response` | **none — never classifies** | 4 |
| `alert_skill.py:455-474` | `tool_called`, or empty response with no tool calls | **none — never classifies** | 6 |
| `metric_skill.py:168-172` | metric created | `"?" in t or "could you" or "please provide" or "clarif"` | 7 |
| `conversation.py:195-199, 330` | turn ends | same but looser (`"please"`), **plus** the structural `chat_result.alert_proposals` signal | 20 |
| `kda_skill.py:236` | create called | `endswith("?")` + the new list branch | **3** |

`visualization` and `alert_skill` — the two closest analogues (run a skill, produce an artifact, may need a round trip) — deliberately never classify the text at all. They break on the *goal signal* (viz created / alert tool called) and on an *empty* response, and otherwise just nudge again. That's exactly the asymmetric-cost reasoning above, and it makes this whole class of bug structurally impossible.

KDA already has the equivalent goal signal (`create_args is not None`, `kda_skill.py:226`). That makes `_is_asking_kda_clarification` a pure *extra early-give-up condition* — dropping it turns KDA into the viz/alert model.

**Suggestion:** rather than adding a second regex rule, consider matching `visualization.py:165-179`: break on `create_args is not None` or empty `response_text`, otherwise always send a simulated reply. That fixes QA-28800 *and* every shape in the table above, with no further tickets. If you keep the classifier, at minimum please widen `_LIST_ITEM_RE` and tolerate a trailing prose line. Either way `_DEFAULT_MAX_ITERATIONS = 3` deserves a bump — `kda_skill.py:18-20` says 3 is already sized exactly for "metric and period each needing their own question", so there is no slack for a wasted turn (viz/alert/metric carry 4–7).

If there is a KDA-specific reason it must classify where viz and alert don't, that's worth a line in the docstring — the current "each skill's heuristic has already drifted independently" reads as license to drift, but the actual drift in this package is only between *loose* and *absent*.

Rest is nits, inline. Nice trace-backed problem statement in the description, and thanks for the period-hint plumbing — that gap was real.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
FrankHuynh: Fixed in `e68ef798` — broadened `_LIST_ITEM_RE` to also match `•`, en-dash `–`, `+`, and a markdown table row (`|...|`), not just `-`/`*`/`1.`/`1)`. Also switched the containing check from `all(...)` to `any(...)` (see my reply on the next comment), which incidentally fixes the wrapped-continuation-line case you flagged too, since it no longer needs every remaining line to match.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
FrankHuynh: Agreed, relaxed as suggested — changed `all(...)` to `any(...)`: one qualifying list line after the `?` is now enough, so a closing sentence like "Let me know which one you'd prefer." no longer sinks the match. Kept the false-positive risk you flagged (rhetorical `?` + unrelated list in a final answer) as an accepted trade-off, per your own reasoning — costs one harmless extra simulated-reply turn, not a wrongly-failed run. Added a direct unit test for this exact shape. Fixed in `e68ef798`.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
FrankHuynh: Fixed — extracted the prompt-building into `_build_clarification_prompt()`, and the "For reference, ..." clause (and each of its two parts) is now omitted entirely when there's nothing usable to reference, instead of asserting "an acceptable metric/fact is None 'None'". Added `test_build_clarification_prompt_omits_reference_clause_when_no_candidates_or_period` to lock this in, plus 2 more direct tests for the other combinations. Fixed in `e68ef798`.

packages/gooddata-eval/src/gooddata_eval/core/agentic/kda_skill.py
FrankHuynh: Agreed, this was the same bug at a narrower scope. Extracted into `_build_period_hint()`, which now builds the hint from whichever of the 3 fields is present (e.g. Date Attribute alone -> "which date dimension?" now gets an answerable hint instead of None). Added 5 direct unit tests covering every field-presence combination. Fixed in `e68ef798`.

packages/gooddata-eval/tests/test_agentic_kda_skill.py
FrankHuynh: Added direct unit tests for `_is_asking_kda_clarification` (the option-list match, and the "list followed by a closing sentence" case), plus `_build_period_hint` and `_build_clarification_prompt`. One deviation from your suggested test: since `all(...)` became `any(...)` per your comment above, "option list followed by prose" is now intentionally `True`, not `False` -- so I added it as a positive regression test instead of the negative one you sketched. Fixed in `e68ef798`.

FrankHuynh:
Thanks for the deep dive — went with the redesign, not another patch.

`_is_asking_kda_clarification` and `_LIST_ITEM_RE` are removed entirely. `_run_once` now matches `visualization.py`/`alert_skill.py`'s own break conditions exactly: break on `create_args is not None` or an empty response, otherwise always send a simulated reply — no text classification at all. Re-ran your 6-row table against this: all 6 now return "always retry" (including `**Option 1**` with no space after the marker), since none of them are inspected anymore.

Also bumped `_DEFAULT_MAX_ITERATIONS` 3 → 4 per your point on the budget having zero slack — matches `visualization.py`'s own budget as the closest analogue.

Left the module docstring's "each skill's heuristic has already drifted independently" line out since there's no more KDA-specific heuristic to justify.

48 tests passing (down from before since `_is_asking_kda_clarification`'s own unit tests went with it), plus a new regression test for the bold-markdown case and direct unit tests for the two helper functions this pulled out (`_build_period_hint`, `_build_clarification_prompt`).

myhoai (approved):

PR #1750, opened 2026-08-21
feat(gooddata-eval): return AgenticEvalOutcome from every agentic kind, not just 3 of 8

Tomkess:
## Summary

`evaluate_agentic_metric_skill`/`evaluate_agentic_alert_skill`/`evaluate_agentic_conversation` return an `AgenticEvalOutcome` (reasoning_steps, conversation_id, response_id) on success and attach the same three attributes to their raised `*AssertionError` on failure. Every other agentic kind — `agentic_guardrail`, `agentic_search`, `agentic_general_question`, `vis_agentic`/`agentic_visualization`, and `agentic_kda_skill` — just returned `None`/a bare value, so their reasoning trace was silently dropped even though `ChatClient`/`_build_chat_result` captures it for every conversation regardless of kind.

This surfaced downstream: a consumer repo switched a `guardrail` fixture from the single-shot evaluator to `agentic_guardrail`, then found a live rerun produced 0/30 `.reasoning.json` sidecars for a kind that previously got them ~14/15 of the time under the single-shot path.

## Changes

- `core/agentic/guardrail.py`, `search_tool.py`, `general_question.py` — same fix, single-message-per-run: capture `chat_result.reasoning_steps`/`.response_id`, attach to the `*Result` dataclass and to the `*AssertionError` on failure, return `AgenticEvalOutcome` on success.
- `core/agentic/visualization.py` — same fix, but multi-turn: accumulates reasoning across every simulated-user iteration in `_execute_single_run`, not just the final turn. Covers both `vis_agentic` and `agentic_visualization`.
- `core/agentic/kda_skill.py` — three-part fix: (1) `agent_id` was never threaded through this kind at all, unlike every other agentic kind; (2) reasoning accumulation, same as above; (3) `cli/agentic_runner.py`'s `agentic_kda_skill` dispatch branch called `evaluate_agentic_kda_skill(...)` without `return` or `agent_id=` — an independent bug that silently dropped its outcome even with (1)+(2) fixed.
- `cli/agentic_runner.py` — `_dispatch_agentic`'s docstring/return type updated to reflect that every kind now returns a real outcome (no more "the rest still return None").

## Tests

- 3 new tests per fixed file (reasoning accumulated on pass, on fail, and across iterations where multi-turn), mirroring the existing `metric_skill` test shape.
- `test_agentic_runner.py`: extended the parametrized dispatch-coverage table to include every kind in `AGENTIC_TEST_KINDS` (was missing `agentic_metric_skill`/`agentic_alert_skill`/`agentic_kda_skill`), added a guard test asserting the table stays in sync with `AGENTIC_TEST_KINDS`, and added a new parametrized test asserting every kind's dispatch returns the *exact* `AgenticEvalOutcome` its evaluator produced. Confirmed this catches regressions by reverting the `kda_skill` `return` fix locally and observing the new test fail, then re-applying it.
- Removed `test_run_agentic_items_defaults_reasoning_steps_to_empty_for_untouched_kinds`, which asserted the old bug (guardrail returning `None`) as intended, permanent behavior.

`uv run pytest packages/gooddata-eval/tests/` — 429 passed (9 pre-existing failures on this branch's base too, all `ModuleNotFoundError: No module named 'openai'` in files this PR doesn't touch — confirmed via `git stash` against unmodified `master`).

## Summary by CodeRabbit

- **New Features**
  - Agent evaluations now return structured outcomes with reasoning steps and response identifiers.
  - Evaluation failures include available reasoning and conversation metadata for improved diagnostics.
  - KDA evaluations support optional agent identification.

- **Tests**
  - Expanded coverage across question, guardrail, search, visualization, KDA, and runner evaluations.
  - Added validation for metadata preservation in successful and failed evaluations.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
@@ -406,6 +406,24 @@ def run_agentic_conversation(
     )
 
 
+def _conversation_detail(result: ConversationResult) -> dict:
+    return {
+        "full_skill_coverage": result.full_skill_coverage,
+        "total_clarification_turns": result.total_clarification_turns,
+        "turns": [
+            {
+                "turn_id": tr.turn_id,
+                "expected_skill": tr.expected_skill,
+                "skill_routing": tr.skill_routing,
+                "output_present": tr.output_present,
+                "output_correct": tr.output_correct,
+                "activated_skills": tr.activated_skills,
+            }
+            for tr in result.turn_results
+        ],
+    }
+
+
 class ConversationAssertionError(AssertionError):
     """Raised when a conversation evaluation fails."""
 
@@ -413,6 +431,7 @@ class ConversationAssertionError(AssertionError):
     reasoning_steps: list[str]
     conversation_id: str
     response_id: str | None
+    detail: dict
 
 
 def evaluate_agentic_conversation(
@@ -524,9 +543,11 @@ def evaluate_agentic_conversation(
         exc.reasoning_steps = result.reasoning_steps
         exc.conversation_id = result.conversation_id
         exc.response_id = result.response_id
+        exc.detail = _conversation_detail(result)
         raise exc
     return AgenticEvalOutcome(
         reasoning_steps=result.reasoning_steps,
         conversation_id=result.conversation_id,
         response_id=result.response_id,
+        detail=_conversation_detail(result),
     )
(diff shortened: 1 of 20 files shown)


packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
```diff
@@ -406,13 +406,32 @@ def run_agentic_conversation(
     )
 
 
+def _conversation_detail(result: ConversationResult) -> dict:
+    return {
```
hkad98, line 410: Nit: no need to fix it now but it would be nice to have a dataclass instead of dict. Feel free to consider it as a follow up.

hkad98 (approved):

PR #1757, opened 2026-08-24
Return the per-turn detail subset from TurnResult

Tomkess:
Follow-up to @hkad98's review comment on #1750 (https://github.com/gooddata/gooddata-python-sdk/pull/1750#discussion_r3841169759): the per-turn subset reported in `detail["turns"]` was built as a bare dict literal inside `_conversation_detail`.

`TurnResult` already owns every field being reported, so the subset is now derived from the model rather than restated:

```python
def detail(self) -> dict:
    return self.model_dump(include=self._DETAIL_FIELDS)
```

`_conversation_detail` becomes `"turns": [tr.detail() for tr in result.turn_results]`. Output shape and the `detail: dict` contract are unchanged.

Two properties come from doing it this way rather than with a hand-written key list:

- A renamed field can't leave `detail()` emitting a stale key silently — `include=` raises on a name the model doesn't have, and a test asserts `_DETAIL_FIELDS` is a subset of `model_fields`.
- `model_dump` deep-copies `activated_skills`, so a caller mutating the returned dict can't reach back into the `TurnResult` ([CodeRabbit finding](https://github.com/gooddata/gooddata-python-sdk/pull/1757#discussion_r3851187704)) — no explicit `list()` wrapper needed.

Verified: 465 passed, including the existing exact-dict assertions on `outcome.detail["turns"]`, unchanged. `ruff check` / `format --check` clean.

**Revision history, since the earlier title/body described something else:** the first attempt added a separate `TurnDetail` dataclass with `asdict()` at the boundary; that was replaced with a method on `TurnResult` per [review](https://github.com/gooddata/gooddata-python-sdk/pull/1757#discussion_r3851037451), and then with `model_dump(include=...)` per the follow-up review — which answers the original "dataclass instead of dict" ask better than either, since the subset is now a derived view of the model instead of a second hand-maintained copy of its field names.

**Ordering note:** #1762 changes what `skill_routing` means relative to `activated_skills`, and this method is what serializes both. #1762 resolves the resulting `{"skill_routing": true, "activated_skills": []}` shape by adding an `active_skills` field carrying the set the credit was drawn from. Whichever lands second should carry that through — if this one lands first, #1762's `TurnResult` change will come back through `_DETAIL_FIELDS`.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 
 import json
 import re
-from dataclasses import asdict, dataclass, field
+from dataclasses import dataclass, field
 from typing import Literal
 
 from gooddata_sdk import GoodDataSdk
@@ -69,6 +69,17 @@ class TurnResult(BaseModel):
     @property
     def skill_success(self) -> bool:
         return self.skill_routing and self.output_present and self.no_error
+
+    def detail(self) -> dict:
+        """The subset of this result reported in detail["turns"] for one conversation turn."""
+        return {
+            "turn_id": self.turn_id,
+            "expected_skill": self.expected_skill,
+            "skill_routing": self.skill_routing,
+            "output_present": self.output_present,
+            "output_correct": self.output_correct,
+            "activated_skills": self.activated_skills,
+        }
 
 
 def _resolve_refs(
@@ -426,35 +437,11 @@ def run_agentic_conversation(
     )
 
 
-@dataclass
-class TurnDetail:
-    """The subset of a TurnResult reported in detail["turns"] for one conversation turn."""
-
-    turn_id: str
-    expected_skill: str
-    skill_routing: bool
-    output_present: bool
-    output_correct: bool | None
-    activated_skills: list[str]
-
-
 def _conversation_detail(result: ConversationResult) -> dict:
     return {
         "full_skill_coverage": result.full_skill_coverage,
         "total_clarification_turns": result.total_clarification_turns,
-        "turns": [
-            asdict(
-                TurnDetail(
-                    turn_id=tr.turn_id,
-                    expected_skill=tr.expected_skill,
-                    skill_routing=tr.skill_routing,
-                    output_present=tr.output_present,
-                    output_correct=tr.output_correct,
-                    activated_skills=tr.activated_skills,
-                )
-            )
-            for tr in result.turn_results
-        ],
+        "turns": [tr.detail() for tr in result.turn_results],
         "latency_breakdown": build_latency_breakdown(result.tool_call_events, result.reasoning_step_events),
     }
 
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
@@ -78,7 +78,7 @@ class TurnResult(BaseModel):
             "skill_routing": self.skill_routing,
             "output_present": self.output_present,
             "output_correct": self.output_correct,
-            "activated_skills": self.activated_skills,
+            "activated_skills": list(self.activated_skills),
         }
 
 
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 import json
 import re
 from dataclasses import dataclass, field
-from typing import Literal
+from typing import ClassVar, Literal
 
 from gooddata_sdk import GoodDataSdk
 from pydantic import BaseModel
@@ -70,16 +70,22 @@ class TurnResult(BaseModel):
     def skill_success(self) -> bool:
         return self.skill_routing and self.output_present and self.no_error
 
+    # Reported per turn in detail["turns"]. A name listed here that no longer exists on the
+    # model raises rather than silently emitting a stale key, which a hand-written dict
+    # literal of the same fields would not -- and model_dump deep-copies activated_skills,
+    # so a caller mutating the returned dict cannot reach back into this TurnResult.
+    _DETAIL_FIELDS: ClassVar[set[str]] = {
+        "turn_id",
+        "expected_skill",
+        "skill_routing",
+        "output_present",
+        "output_correct",
+        "activated_skills",
+    }
+
     def detail(self) -> dict:
         """The subset of this result reported in detail["turns"] for one conversation turn."""
-        return {
-            "turn_id": self.turn_id,
-            "expected_skill": self.expected_skill,
-            "skill_routing": self.skill_routing,
-            "output_present": self.output_present,
-            "output_correct": self.output_correct,
-            "activated_skills": list(self.activated_skills),
-        }
+        return self.model_dump(include=self._DETAIL_FIELDS)
 
 
 def _resolve_refs(
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 
 import json
 import re
-from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
```diff
@@ -393,19 +393,33 @@ def run_agentic_conversation(
...
+                TurnDetail(
+                    turn_id=tr.turn_id,
+                    expected_skill=tr.expected_skill,
+                    skill_routing=tr.skill_routing,
+                    output_present=tr.output_present,
+                    output_correct=tr.output_correct,
+                    activated_skills=tr.activated_skills,
+                )
+            )
             for tr in result.turn_results
```
hkad98: Proposal: the output is a dict. Consider adding a method called detail() which output will be a dictionary and will be part of TurnResult.

Tomkess: Good call — done in 4442940. Dropped the separate `TurnDetail` dataclass entirely and added `TurnResult.detail()` instead, since `TurnResult` already owns every field being reported. `_conversation_detail` now just does `[tr.detail() for tr in result.turn_results]`. Output shape unchanged — all 19 `test_agentic_conversation.py` tests (including the exact-dict assertions on `detail["turns"]`) still pass as-is.

packages/gooddata-eval/src/gooddata_eval/core/agentic/conversation.py
```diff
@@ -64,6 +64,17 @@ class TurnResult(BaseModel):
...
+    def detail(self) -> dict:
+        """The subset of this result reported in detail["turns"] for one conversation turn."""
+        return {
+            "turn_id": self.turn_id,
+            "expected_skill": self.expected_skill,
+            "skill_routing": self.skill_routing,
+            "output_present": self.output_present,
+            "output_correct": self.output_correct,
+            "activated_skills": self.activated_skills,
+        }
```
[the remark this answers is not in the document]
Tomkess: Fixed in 9aa8d86 — `detail()` now returns `list(self.activated_skills)`, plus a regression test asserting a mutation on the returned dict's list doesn't leak back into the TurnResult.

hkad98:
The diff doesn't do what the title and body say it does.

The body states it "Adds `TurnDetail` (a small `@dataclass`) mirroring the 6 fields previously built as a bare dict literal" with "`asdict()` at the boundary". There is no dataclass and no `asdict()` in the diff — what landed is a `TurnResult.detail()` method returning the same dict literal, relocated from `_conversation_detail` into the model:

```python
def detail(self) -> dict:
    return {
        "turn_id": self.turn_id,
        ...
    }
```

So the review ask this follows up on ("it would be nice to have a dataclass instead of dict") isn't addressed — the dict literal moved, it didn't become a type. Either the branch was reworked and the description is stale, or the description describes intent rather than the change. Worth reconciling before merge, since the title is what ends up in the commit log.

On the substance: `TurnResult` is already a pydantic `BaseModel`, so a hand-maintained key list is redundant regardless of which direction you go:

```python
def detail(self) -> dict:
    return self.model_dump(
        include={
            "turn_id",
            "expected_skill",
            "skill_routing",
            "output_present",
            "output_correct",
            "activated_skills",
        }
    )
```

That deep-copies `activated_skills` for free, so the new regression test still passes, and it can't drift when a field is renamed — the current version would keep emitting a stale key name with no failure. Key ordering differs from the literal, but the tests compare dicts, so that's not observable.

Two nits:

- The body carries a "Test plan" section and a `Generated with Claude Code` line; neither belongs in our PR bodies.
- `test_turn_result_detail_returns_independent_activated_skills_list` is 71 characters and duplicates its own docstring. `test_turn_result_detail_copies_activated_skills` says the same thing.

One cross-PR note: #1762 makes `skill_routing` cumulative across turns while leaving `activated_skills` per-turn. Whichever of these merges second, this method is the thing that serializes both fields side by side, so it'll be where the resulting `{"skill_routing": true, "activated_skills": []}` shows up in reports.

hkad98:
Nothing has changed on this one — no new commits since my comment, and the body still describes a `TurnDetail` dataclass with `asdict()` that the diff doesn't contain (`Test plan` section and the `Generated with Claude Code` line are also still there). Flagging in case it got lost while #1762 and #1772 were being worked on, rather than deliberately parked.

Still the smallest of the three to close out. Either direction works:

- **Keep the method, drop the hand-written keys** — `self.model_dump(include={...})`, then retitle to match ("return the per-turn detail subset from `TurnResult`" or similar).
- **Actually add the dataclass** the title promises, if the point is to give `detail["turns"]` a named type.

The first is fewer moving parts and still answers @hkad98's original ask better than the current diff does, since the subset becomes a derived view of the model rather than a second hand-maintained copy of its field names.

Worth noting the ordering constraint with the other two: #1762 will change what `skill_routing` means relative to `activated_skills`, and this method is what serializes both. If #1762 lands first, whatever you do here should carry its resolution of the `{"skill_routing": true, "activated_skills": []}` shape. If this lands first, that fix will have to come back through here.

Tomkess:
@hkad98 you're right on all of it, and apologies for the delay — your two comments here sat unread while I was working #1762/#1772, exactly the "lost rather than deliberately parked" case you guessed at.

**Title/body vs. diff** — reconciled. The title now describes what the diff does (`Return the per-turn detail subset from TurnResult`), and the body records the revision history explicitly so the mismatch isn't confusing in the log. The `Test plan` section and the `Generated with Claude Code` line are gone; noted for future PRs.

**`model_dump(include=...)`** — taken, in `0aa2b77e`. You're right that a hand-written list of the model's own field names was the same duplication problem one layer over, and it does answer the original ask better: the subset is now a derived view of the model rather than a second copy. Two things fell out of it that I hadn't expected:

- The deep copy comes for free, so the explicit `list(self.activated_skills)` wrapper is gone and the copy regression test still passes.
- I kept the field names in a `_DETAIL_FIELDS` ClassVar and added a test asserting it's a subset of `model_fields`, so a rename fails loudly instead of dropping a key from every report. That's the drift you flagged, now covered.

**Nits** — test renamed to `test_turn_result_detail_copies_activated_skills`, and the duplicated construction lifted into a helper.

**Cross-PR ordering** — agreed, and #1762 has since resolved that shape: it adds `TurnResult.active_skills` carrying the set `skill_routing` was judged against, so `{"skill_routing": true, "activated_skills": []}` is self-explanatory rather than looking like a scoring bug. Noted in the body here too. If this lands first, that field comes back through `_DETAIL_FIELDS`.

Also worth flagging from the #1762 thread since it touches this file: the real `set_skills` argument key is `skill_names`, and our conversation tests were using a bare `skills` — so they'd been passing through `_activated_skills`' fallback rather than the real payload shape. Fixed on that branch.

465 passed, lint clean.

hkad98 (approved):
Approving — title, body and diff now agree, and `model_dump(include=...)` is the right shape. 465 passed on my end too. Recording the revision history in the body was the right way to handle the mismatch.

One factual correction, because it's a claim about a safety property that doesn't exist. Both the body and the inline comment say:

> A renamed field can't leave `detail()` emitting a stale key silently — `include=` raises on a name the model doesn't have

`include=` does **not** raise. Pydantic v2 silently ignores names in `include` that aren't fields:

```python
class M(BaseModel):
    a: int
    b: str
    _F: ClassVar[set[str]] = {"a", "b", "renamed_away"}

M(a=1, b="x").model_dump(include=M._F)
# -> {'a': 1, 'b': 'x'}      no error, key just absent
```

So the drift protection here comes entirely from your test, not from `model_dump`:

```python
assert set(TurnResult.model_fields) >= TurnResult._DETAIL_FIELDS
assert set(_turn_result().detail()) == TurnResult._DETAIL_FIELDS
```

Both catch it — the second is the stronger of the two, since it fails on the actual output rather than on the declaration. That's a real guarantee and it's sufficient; the only problem is the comment crediting the wrong mechanism, which would leave a future reader thinking they'd still be covered if they deleted the test. Worth rewording the `_DETAIL_FIELDS` comment and the body bullet to point at the test instead.

`_DETAIL_FIELDS: ClassVar[set[str]]` is correctly annotated, so pydantic treats it as a class attribute rather than a field — worth noting since an unannotated `_`-prefixed attribute on a `BaseModel` would have behaved differently.

No need to re-request review for the comment fix; approving as-is.

PR #1760, opened 2026-08-26
fix(gooddata-eval): scope simulated-user pushback to the original request

FrankHuynh:
## Summary
- `generate_simulated_response()` only received the assistant's last message and the ground-truth MAQL, and its prompt forced every clause of that MAQL to be satisfied "even if the assistant's question doesn't explicitly ask about it" — so the simulated user would inject filters/constraints the original request never mentioned, even when the assistant's proposal already matched it.
- Threaded the original question through (`_execute_single_metric_run` already has `question` in scope; `conversation.py`'s `TurnDefinition.message` carries the same for multi-turn `agent_conversations`), and rewrote the prompt to agree when the original request is already satisfied, only adding a clause when it's a reasonable reading of that request.

## Test plan
- [x] `pytest packages/gooddata-eval/tests/test_agentic_metric_skill.py packages/gooddata-eval/tests/test_agentic_conversation.py -q` — 50 passed
- [x] Full `gooddata-eval` suite — 464 passed
- [x] `ruff check` on changed files — clean (3 pre-existing unrelated errors elsewhere in the package, untouched by this change)
- [x] Added a regression test (`test_generate_simulated_response_prompt_agrees_when_the_original_request_is_already_satisfied`) reproducing the QA-29094 scenario (chatbot proposal already satisfies the original ask → simulated user must agree, not inject a new filter)
- [x] Updated the existing `test_generate_simulated_response_prompt_preserves_maql_fidelity` assertions to match the new conditional prompt wording (dropped the now-removed "every clause" unconditional phrase, kept the verbatim/filter-fidelity checks it originally guarded)

jira: QA-29094
risk: nonprod

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metric clarification conversations by retaining the original request when evaluating proposed answers.
  * Reduced unnecessary clarification when a proposal already satisfies the request.
  * Preserved relevant metric filters, including those in quoted text and identifiers.
  * Improved error reporting during metric conversations with clearer diagnostic details.

* **Tests**
  * Expanded coverage for clarification behavior, filter detection, original-request context, and error scenarios.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
@@ -30,6 +30,11 @@ _INNER_SELECT_RE = re.compile(r"\(\s*SELECT\s*\{([^}]+)\}\s*\)", re.IGNORECASE)
 # Everything else in MAQL (keywords, operators, numbers, punctuation) carries no
 # case-sensitive meaning, per the MAQL reference (SELECT/BY/WHERE/FOR PREVIOUS/etc.
 # are case-insensitive; only {..} identifiers and quoted literal values are not).
+# Feeds _normalize_maql, the scoring comparator (_best_maql_match) -- do not widen this
+# to handle \X escapes without confirming MAQL literals actually support backslash
+# escaping (unconfirmed; see PR #1760 review). A wrong guess here silently changes
+# maql_correct for the whole eval dataset, not just a hint. _no_where_clause_hint()
+# below has its own, separately-scoped regex for that reason.
 _PROTECTED_RE = re.compile(r"\{[^}]*\}|\"[^\"]*\"|'[^']*'")
 
 
@@ -99,8 +104,42 @@ class SimulatedResponseError(RuntimeError):
     """
 
 
-def generate_simulated_response(agent_message: str, expected_output: dict) -> str:
+# Separate from _PROTECTED_RE on purpose: this one only feeds a same-turn LLM-prompt hint
+# (see _no_where_clause_hint), never the scoring comparator, so it can afford to consume
+# \X escape sequences inside quoted literals without risking maql_correct semantics.
+_HINT_PROTECTED_RE = re.compile(r"\{[^}]*\}|\"(?:[^\"\\]|\\.)*\"|'(?:[^'\\]|\\.)*'")
+
+
+def _no_where_clause_hint(expected_maqls: list[str]) -> str:
+    """Deterministic nudge for when NONE of the accepted candidate MAQLs has a WHERE clause.
+
+    Without this, whether to add a filter is left entirely to the simulating LLM's judgment
+    of what the original request "implies" -- the same fuzzy reasoning that caused it to
+    inject an unrequested filter in the first place (QA-29094). Checks every candidate, not
+    just the first: _best_maql_match accepts any of them, so hinting off just candidate 0
+    would risk steering the agent away from a filtered candidate the scorer would still have
+    accepted (the mirror-image of the original bug). Strips {type/id} identifiers and quoted
+    literals first so a "where" substring inside one of those -- e.g.
+    `{metric/somewhere_sales}`, or a literal value containing the word -- doesn't get
+    mistaken for a real WHERE clause.
+    """
+    for maql in expected_maqls:
+        outside_protected = _HINT_PROTECTED_RE.sub(" ", maql)
+        if re.search(r"\bWHERE\b", outside_protected, re.IGNORECASE):
+            return ""
+    return (
+        " This metric needs no filter. If the assistant asks about excluding or filtering "
+        "anything, say no filter is needed."
+    )
+
+
+def generate_simulated_response(agent_message: str, expected_outputs: list[dict], original_question: str) -> str:
     """Generate a user reply to keep the metric-skill conversation going (gpt-4o-mini).
+
+    ``expected_outputs`` is the fixture's full candidate list (as accepted by
+    ``_best_maql_match``), not just the first one -- the ground-truth MAQL woven into the
+    prompt still comes from candidate 0, but the no-filter hint checks all of them (see
+    ``_no_where_clause_hint``).
 
     Raises:
         SimulatedResponseError: openai is not installed, OPENAI_API_KEY is unset, or the
@@ -116,15 +155,23 @@ def generate_simulated_response(agent_message: str, expected_output: dict) -> st
         raise SimulatedResponseError("OPENAI_API_KEY environment variable is not set")
 
     client = OpenAI(api_key=api_key)
-    expected_maql = expected_output.get("maql", "")
+    expected_maql = expected_outputs[0].get("maql", "") if expected_outputs else ""
+    expected_maqls = [eo.get("maql", "") for eo in expected_outputs]
     prompt = (
         f"You are simulating a user in a conversation with a BI assistant that creates metrics. "
+        f"The user's original request was: '{original_question}'. "
         f"The assistant said: '{agent_message}'. "
         f"The user's ground-truth intended metric is exactly this MAQL: {expected_maql}. "
-        f"Reply as the user. You MUST ensure every clause of that MAQL (including any WHERE/filter "
-        f"conditions) is eventually satisfied, and quote field/label identifiers verbatim from it -- "
-        f"never paraphrase or drop a clause, even if the assistant's question doesn't explicitly ask "
-        f"about it. If the assistant's offered options omit a required filter, add it yourself."
+        f"Reply as the user. If the assistant is asking a clarifying question rather than proposing "
+        f"a metric, answer that question directly using the ground-truth MAQL -- quote field/label "
+        f"identifiers verbatim -- instead of merely agreeing. "
+        f"If the assistant's proposal already satisfies the ORIGINAL REQUEST above, agree and confirm "
+        f"-- do not introduce new requirements the original request never mentioned. "
+        f"Only if the assistant's proposal is missing something the original request actually implies "
+        f"(e.g. a filter/clause from the ground-truth MAQL that is a reasonable reading of the original "
+        f"request), point it out and add it yourself, quoting field/label identifiers verbatim from the "
+        f"ground-truth MAQL."
+        f"{_no_where_clause_hint(expected_maqls)}"
     )
     try:
         response = client.chat.completions.create(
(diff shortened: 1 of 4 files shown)


FrankHuynh:
Addressed all 5 findings from review:

**P1-1 (clarifying question branch missing)** — added an explicit prompt branch: if the assistant is asking a clarifying question (no proposal yet), the simulated user now answers directly from the ground-truth MAQL instead of trivially agreeing.

**P1-2 (no deterministic no-filter guard)** — extracted `_no_filter_hint()`: when the ground-truth MAQL has no `WHERE` clause, the prompt now explicitly tells the simulated user no filter is needed, instead of relying on the LLM's judgment of what the original request "implies" (the same fuzzy reasoning that caused the original bug).

**P2-1 (uncovered call site + silent except)** — added a direct unit test for `_get_sim_user_response`'s metric branch (previously 0% coverage, confirmed via `--cov`), and changed `except Exception: pass` to log the exception instead of swallowing it.

**P2-2 (dropped max_tokens assertion)** — restored `assert call_kwargs["max_tokens"] >= 300` in the fidelity test.

**P2-3 (brittle prose assertions)** — reduced the new tests' reliance on exact prompt wording; they now mostly check interpolated data (original_question/expected_maql presence) and the independently-testable `_no_filter_hint()` output rather than full sentences.

468 tests passing (was 464), ruff clean, coverage gap on `conversation.py`'s metric branch closed.

packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
```diff
@@ -99,7 +99,22 @@ class SimulatedResponseError(RuntimeError):
...
 
-def generate_simulated_response(agent_message: str, expected_output: dict) -> str:
+def _no_filter_hint(expected_maql: str) -> str:
+    """Deterministic nudge for when the ground-truth MAQL has no WHERE clause.
+
+    Without this, whether to add a filter is left entirely to the simulating LLM's judgment
+    of what the original request "implies" -- the same fuzzy reasoning that caused it to
+    inject an unrequested filter in the first place (QA-29094).
+    """
+    if "WHERE" in expected_maql.upper():
```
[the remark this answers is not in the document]
FrankHuynh: Fixed — `_no_filter_hint()` now strips `{type/id}` identifiers and quoted literals via `_PROTECTED_RE` before matching `\bWHERE\b`, so `{metric/somewhere_sales}` no longer false-positives. Added regression tests for an identifier and a literal both containing "where". See 8cbca3cf.

packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
```diff
@@ -99,7 +99,26 @@ class SimulatedResponseError(RuntimeError):
...
+
+    Without this, whether to add a filter is left entirely to the simulating LLM's judgment
+    of what the original request "implies" -- the same fuzzy reasoning that caused it to
+    inject an unrequested filter in the first place (QA-29094). Strips {type/id} identifiers
+    and quoted literals first (same protected-span rule as `_casefold_outside_protected`) so
+    a "where" substring inside one of those -- e.g. `{metric/somewhere_sales}`, or a literal
+    value containing the word -- doesn't get mistaken for a real WHERE clause.
+    """
+    outside_protected = _PROTECTED_RE.sub(" ", expected_maql)
+    if re.search(r"\bWHERE\b", outside_protected, re.IGNORECASE):
```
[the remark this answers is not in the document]
FrankHuynh: Fixed — `_PROTECTED_RE`'s quoted-literal alternatives now consume \X escape sequences (`"(?:[^"\\\\]|\\\\.)*"`) so an escaped quote no longer ends the match early and leaks the rest of the literal (including a standalone WHERE) as unprotected. Added a regression test with an escaped quote followed by WHERE inside a literal. See 8cbca3cf.

packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
```diff
@@ -30,7 +30,10 @@
 # Everything else in MAQL (keywords, operators, numbers, punctuation) carries no
 # case-sensitive meaning, per the MAQL reference (SELECT/BY/WHERE/FOR PREVIOUS/etc.
 # are case-insensitive; only {..} identifiers and quoted literal values are not).
-_PROTECTED_RE = re.compile(r"\{[^}]*\}|\"[^\"]*\"|'[^']*'")
+# The quoted-literal alternatives consume \X escape sequences (including an escaped
+# quote) so an escaped quote inside the literal doesn't end the match early and leak
+# the rest of the literal's text as unprotected.
+_PROTECTED_RE = re.compile(r"\{[^}]*\}|\"(?:[^\"\\]|\\.)*\"|'(?:[^'\\]|\\.)*'")
```
henrynguyengooddata: This regex change is out of scope for QA-29094 and lands in the scoring path, not in the new hint.

`_PROTECTED_RE` is shared: `_casefold_outside_protected` (L54–59) → `_normalize_maql` (L67) → `_best_maql_match` (L84) — i.e. the comparator that decides `maql_correct`. So this edit changes pass/fail semantics for MAQL comparison, while the four new tests only cover `_no_filter_hint`. There is no test on `_normalize_maql` for the input the regex was changed for.

Behaviour does diverge. Repro (old vs new regex through `_casefold_outside_protected`):

```
IN  : SELECT {metric/x} WHERE {label/path} = "C:\" AND {label/y} = "Active"
OLD : select {metric/x} where {label/path} = "C:\" and {label/y} = "Active"
NEW : select {metric/x} where {label/path} = "C:\" AND {label/y} = "active"
                                                    ^^^ not casefolded  ^^^ literal casefolded
```

The trigger is narrow — a literal ending in an odd run of backslashes — so I'm not claiming this fires often. The concern is that it changes the comparator at all, in a direction we haven't established:

- If MAQL string literals **do not** support `\` escapes (a backslash is just a character), the old regex was right and the new one over-consumes past the closing quote.
- If they **do**, the old regex was wrong and this is a real fix — but then it deserves its own commit, a `_normalize_maql` test, and a note in the ticket.

Right now the PR assumes the second without evidence, and either way the change isn't tested.

Could we do one of these?

1. **Keep QA-29094 scoped**: give `_no_filter_hint` its own local regex and leave `_PROTECTED_RE` untouched. The hint only needs to blank out spans before a `\bWHERE\b` search, so it can afford to be conservative in a way the comparator cannot.
2. **Keep the change**, but confirm MAQL's escaping rule (docs or backend parser), add `_normalize_maql` cases for `"…\"` and `'…\'`, and say so in the PR description so it's reviewed as a comparator change rather than a helper detail.

I'd lean towards (1) for this PR and (2) as a follow-up, so a flaky-test fix can't move the pass/fail line underneath the eval datasets.

packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
```diff
@@ -99,7 +102,26 @@ class SimulatedResponseError(RuntimeError):
     """
 
 
-def generate_simulated_response(agent_message: str, expected_output: dict) -> str:
+def _no_filter_hint(expected_maql: str) -> str:
```
henrynguyengooddata: `_no_filter_hint` only ever sees the first candidate, which can turn QA-29094 into the mirror-image false failure.

`_execute_single_metric_run` passes `primary_expected = expected_outputs[0]` (L266 → L313), so both the prompt's ground-truth MAQL and this hint are derived from candidate 0 alone. But `_best_maql_match` (L84) accepts **any** candidate, and the ticket's own root cause calls this out:

> a single "ground-truth" MAQL (expected_outputs[0], the first of possibly several valid fixture candidates)

So for a fixture whose candidate 0 has no `WHERE` but candidate 1 does, the simulated user is now told, deterministically:

> say no filter is needed

That actively steers the agent away from a candidate the scorer would have accepted. The old bug was "injects a filter nobody asked for"; this is "blocks a filter the fixture allows". Same false `maql_correct=0`, opposite direction — and harder to spot, because the hint reads as the safe choice.

Suggestion — take the whole candidate list and only emit the hint when *none* of them filter:

```python
def _no_filter_hint(expected_maqls: list[str]) -> str:
    if any(
        re.search(r"\bWHERE\b", _PROTECTED_RE.sub(" ", maql), re.IGNORECASE)
        for maql in expected_maqls
    ):
        return ""
    return " This metric needs no filter. …"
```

That means threading `expected_outputs` (not just `primary_expected`) into `generate_simulated_response`. It's a slightly bigger signature change, but it's the part of the stated root cause the PR currently leaves open, and it's cheaper to do now than after the prompt ships.

Two smaller things while we're here:

- **Naming**: the function says "no *filter*" but only matches `\bWHERE\b`. Either rename it `_no_where_clause_hint` or widen the detection — right now the name promises more than the code checks.
- **Hardcoded example**: `"(e.g. cancelled orders, a status, a date range)"` lifts the QA-29094 repro domain into a generic helper. Every metric fixture now gets "cancelled orders" in its prompt, which is exactly the kind of nudge we're trying to remove. Suggest dropping the concrete example.

henrynguyengooddata (commented):
Two comments on the QA-29094 fix — the direction looks right and the new test coverage (especially `test_get_sim_user_response_metric_branch_forwards_the_turn_message`) is a solid addition. Both points below are about scope and about the part of the stated root cause that is still open.

packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
FrankHuynh: Good catch, fixed — reverted `_PROTECTED_RE` to the original non-escape-aware pattern (so the comparator's behavior is unchanged from before this PR), and gave `_no_where_clause_hint` its own separately-scoped `_HINT_PROTECTED_RE` for the escape-handling. Added `test_normalize_maql_does_not_consume_escape_sequences` to lock in the comparator's original behavior with your exact repro case. Agree this needs its own investigation + normalize_maql tests before touching the shared regex — filing that as a follow-up rather than in this PR. See a7da5561.

packages/gooddata-eval/src/gooddata_eval/core/agentic/metric_skill.py
FrankHuynh: Fixed both main points and the two smaller ones:
- `_no_where_clause_hint` now takes the full `expected_outputs` candidate list (threaded through `generate_simulated_response`) and only fires when *none* of the candidates has a WHERE clause -- exactly your suggested fix. Added `test_no_where_clause_hint_stays_silent_if_any_candidate_has_a_where_clause` using the actual agent_metric_skill_4.json candidates (unfiltered + Processed-filtered) as the regression case.
- Renamed `_no_filter_hint` -> `_no_where_clause_hint` to match what it actually checks.
- Dropped the hardcoded "(e.g. cancelled orders, a status, a date range)" example from the hint text.

See a7da5561.

henrynguyengooddata (approved):

PR #1763, opened 2026-08-27
feat(gooddata-eval): register knowledge_question evaluator

Tomkess:
## Summary
- `knowledge_question` fixtures were reclassified out of `general_question` (platform/product/policy Q&A that isn't LDM-grounded) but no evaluator was ever registered for the kind, so `get_evaluator("knowledge_question")` raised `KeyError` and every item was silently `skipped` — the downstream repo has kept `knowledge_question: false` in its test-kind config as a result.
- Registers `knowledge_question` in the lazy-evaluator registry, reusing `GeneralQuestionEvaluator` directly rather than adding a new class: both are free-text-rubric, LLM-judged prose answers, differing only in subject matter (LDM facts vs. platform/policy facts).
- `ItemReport.test_kind` is tagged from the dataset item's own `test_kind` field (`runner.py`), not from the evaluator class's `test_kind` attribute, so sharing one evaluator class across both kinds does not mislabel results in reports/dashboards.

## Review follow-ups
Two of @hkad98's three nits are addressed:
- `GeneralQuestionEvaluator.test_kind` is now inaccurate for half its registrations — added a pointer to the registry comment explaining why it's harmless, so nothing derives an output label from it unaware.
- `supported_test_kinds()`'s docstring listed only `(general_question, guardrail)`; now lists all four LLM-judge kinds.

The third (no `agentic_knowledge_question` dispatch branch, so the kind is single-turn only) is left as scope per the review — likely its own follow-up.

## Test plan
- [x] `uv run pytest packages/gooddata-eval/tests/` — 702 passed (rebased onto #1771, which grew the suite)
- [x] New coverage: `get_evaluator("knowledge_question")` resolves to `GeneralQuestionEvaluator`; pass/fail dispatch through the reused evaluator (`test_text_evaluators.py`)
- [x] Updated `test_run_items_routes_all_supported_kinds`'s exact-set assertion to include `knowledge_question`
- [x] `ruff check` / `ruff format --check` clean

## Rebase note
Rebased onto `45892f79` after #1771 landed and put this into conflict. The only clash was an import collision in `test_text_evaluators.py` (#1771 added `JudgeResponseError`, this PR added `get_evaluator`) — both kept, nothing semantic.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/evaluators/__init__.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/evaluators/__init__.py
@@ -20,18 +20,26 @@ _EAGER_EVALUATORS: dict[str, Evaluator] = {
     )
 }
 
-# LLM-judge evaluators (general_question, guardrail, dashboard_summary) require the
-# [llm-judge] extra. Their modules are imported lazily on first use so the CLI
+# LLM-judge evaluators (general_question, guardrail, dashboard_summary, knowledge_question)
+# require the [llm-judge] extra. Their modules are imported lazily on first use so the CLI
 # starts without openai.
+#
+# knowledge_question reuses GeneralQuestionEvaluator directly: both are free-text-rubric,
+# LLM-judged prose answers -- knowledge_question just covers platform/product/policy facts
+# instead of LDM-grounded ones. `ItemReport.test_kind` is tagged from the dataset item's own
+# `test_kind` field (see runner.py), not from the evaluator class, so sharing one class
+# across both kinds does not mislabel results.
 _LAZY_EVALUATOR_MODULES: dict[str, str] = {
     "general_question": "gooddata_eval.core.evaluators.general_question",
     "guardrail": "gooddata_eval.core.evaluators.guardrail",
     "dashboard_summary": "gooddata_eval.core.evaluators.summary",
+    "knowledge_question": "gooddata_eval.core.evaluators.general_question",
 }
 _LAZY_EVALUATOR_CLASSES: dict[str, str] = {
     "general_question": "GeneralQuestionEvaluator",
     "guardrail": "GuardrailEvaluator",
     "dashboard_summary": "DashboardSummaryEvaluator",
+    "knowledge_question": "GeneralQuestionEvaluator",
 }
 
 
--- a/packages/gooddata-eval/src/gooddata_eval/core/evaluators/__init__.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/evaluators/__init__.py
@@ -65,9 +65,9 @@ def _openai_available() -> bool:
 def supported_test_kinds() -> set[str]:
     """Return all supported test_kind values.
 
-    LLM-judge kinds (general_question, guardrail) are excluded when the
-    [llm-judge] extra (openai) is not installed — those items are skipped
-    rather than erroring out mid-run.
+    LLM-judge kinds (general_question, guardrail, dashboard_summary,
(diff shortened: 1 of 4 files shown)


hkad98 (dismissed):
Approving.

Verified the two load-bearing claims in the new comment:

- `ItemReport.test_kind` is populated from `item.test_kind` (`runner.py:112`), not from the evaluator class, so registering one class under two keys does not mislabel results.
- `GeneralQuestionEvaluator`'s rubric (`_EVALUATION_STEPS`) is generic free-text scoring with no LDM- or workspace-specific wording, so it genuinely transfers to platform/product/policy facts.

The lazy-registration path keys off the explicit `_LAZY_EVALUATOR_*` dicts rather than `ev.test_kind`, so the double registration is safe as written.

Three non-blocking nits for whenever you next touch this file:

1. `GeneralQuestionEvaluator.test_kind = "general_question"` is now inaccurate for half its registrations. Harmless today, but it would mislabel if anything ever derives output labels from `evaluator.test_kind` instead of `item.test_kind`. Worth a pointer from the class to the comment you added here.
2. There is no `agentic_knowledge_question` branch in `_dispatch_agentic` (`cli/agentic_runner.py`), so the kind is single-turn only. Fine as scope, just flagging it as the likely next request.
3. Pre-existing: `supported_test_kinds()`'s docstring still lists only "(general_question, guardrail)" and is now two kinds out of date.

Tomkess:
@hkad98 thanks for the approval and the nits — fixed the first two in 068c393b:

1. Added a comment on `GeneralQuestionEvaluator.test_kind` pointing to the registry comment explaining it's now shared by two kinds, so nothing derives a label from it unaware.
2. `supported_test_kinds()`'s docstring now lists all 4 LLM-judge kinds (`general_question`, `guardrail`, `dashboard_summary`, `knowledge_question`) instead of the stale 2.

Left the third (`agentic_knowledge_question` dispatch) out — per your own note, that's scope for a future PR, not this one. 476 passed, lint clean.

Tomkess:
Heads-up: #1771 landing put this into conflict after your approval. Rebased onto `45892f79` as `f2133312` — `mergeable_state` is back to `blocked` rather than `dirty`.

The only conflict was an import collision in `test_text_evaluators.py` (#1771 added `JudgeResponseError`, this PR added `get_evaluator`) — both kept, nothing semantic. The registry change itself didn't conflict.

Re-verified after the rebase: 702 passed (suite grew from 476 with #1771), `ruff check`/`format --check` clean. Your two nit fixes from `068c393b` are carried through unchanged.

hkad98 (approved):
Re-approving — my earlier approval was dismissed by the rebase, not by anything wrong with the PR.

Both nits from `068c393b` are in and are the right shape. The comment on `GeneralQuestionEvaluator.test_kind` says exactly the useful thing (harmless today, but don't derive a label from it without accounting for the second kind), and `supported_test_kinds()`'s docstring now lists all four LLM-judge kinds. Agreed on leaving `agentic_knowledge_question` out — that was my note as future scope, not a request.

Your read on the rebase conflict matches mine: `test_text_evaluators.py` import collision only (#1771's `JudgeResponseError` vs this PR's `get_evaluator`), both kept, nothing semantic. The registry change itself was untouched.

Verified against **current** master (`fb6852d6`, which now has #1757, #1762, #1772 and #1775 on top of the `45892f79` you rebased onto):

| Check | Result |
|---|---|
| Merges into current master | clean |
| Suite, merged | **713 passed** |
| `ruff check` / `format --check` | clean, 85 files |
| `ty check` | All checks passed |

Nothing further from me.

PR #1771, opened 2026-09-02
perf: take Langfuse trace linking off the eval item critical path

tychtjan:
## What and why

Trace linking used to happen inline, per item: the run waited for Langfuse to ingest each
conversation's trace before starting the next item. That wait decides nothing — the pass/fail
verdict is already final by the time it begins — and it was charged to the item, so the agent
looked about twice as slow as it really is.

Those lookups are now queued and run as one batch after the agent phase, before any report is
printed, so **scores are still final before the command exits**. `--concurrency` also reaches
the agentic kinds now. Direct library callers (the tavern e2e suite) keep the old synchronous
behaviour.

Measured on staging — 18 `agentic_general_question` items × 2 runs, `gpt-5.6-luna`:

| | master | this PR |
|---|---|---|
| `--concurrency 1` | 462s | **233s** (2.0× faster) |
| `--concurrency 2` | 774s ¹ | **114s** (4.1× faster) |
| reported latency per run | 12.7s | **5.2s** ² |
| traces lost (scores orphaned) | 0, then 2 ³ | 0 |

¹ `--concurrency` has no effect on master's agentic path, so this is the same serial work as the
row above. Master ran identical workloads in 462s and 774s — inline polling made the runtime
hostage to whatever the ingestion lag happened to be.
² Not a speedup. Master's per-run number *included* that run's Langfuse polling; the agent was
never 12.7s/run. Two separate wins: real wall-clock time, and a latency figure that finally
measures the agent instead of our own bookkeeping.
³ One master run lost two traces. Concurrency overlaps each item's query window, and without a
server-side `sessionId` filter a busy window pushes the item's own trace off the results page.

## Which datasets run in parallel, and which deliberately do not

`--concurrency K` evaluates K items at once. Kinds whose agent **creates objects in the
workspace** are always run one at a time, whatever `K` says:

| runs in parallel | always serial (creates workspace objects) |
|---|---|
| `agentic_general_question` | `agentic_metric_skill` |
| `agentic_guardrail` | `agentic_alert_skill` |
| `agentic_search` | `agentic_kda_skill` |
| `vis_agentic`, `agentic_visualization` | `agentic_conversation` |

The reason is contamination, not speed: a metric or alert created and dropped mid-run would be
visible to another item reading the same catalog, so those items would start failing for
reasons that have nothing to do with what they test. Anything not on the parallel-safe list
runs serially by default, so a kind added later is safe until someone reviews it.

Two caveats: the **single-turn** `metric_skill` / `alert_skill` kinds are still fanned out by
`--concurrency` (the allowlist only guards the agentic dispatch), so avoid raising `K` on a
dataset of those against a shared workspace. And with `K > 1` the parallel-safe kinds all run
before the serial ones, so execution order differs from `--concurrency 1`.

## Behaviour changes reviewers should know

- **Pass rates are not directly comparable to older runs.** A judge fault on one run no longer
  errors the whole item: that run is ungraded, `pass@K` holds on the runs that were graded, and
  `unscored_runs` appears in the detail. An item with no graded run at all errors. `pass^K` is
  stricter — it now requires every run to be graded.
- **`agentic_conversation` no longer claims K runs.** It takes no `k` and drives its fixture
  once, so `--runs 5` used to report five runs and divide one conversation's latency by five.
- **Each item now reports `runs_passed` / `pass_power_k`**; previously a 5/5 item and a 1/5 item
  looked identical in every field.
- **A malformed judge score raises instead of scoring 0.** `{"score": 2}` used to be reported as
  a confident FAIL. `{"score": "1"}` is accepted again.
- **`openai>=1.45`** is required by the `llm-judge` extra — 1.44 and earlier have no
  `max_completion_tokens`, so every judge call would `TypeError`.
- **Ctrl-C now stops promptly.** Queued items and queued trace lookups are dropped, and a trace
  poll that is already running is told to stop rather than sleeping out the rest of its retry
  budget — measured against staging with Langfuse unreachable, the CLI exited **105.4s** after
  the interrupt before this change and **1.1s** after. Agent calls already in flight still
  finish, so expect to wait up to one `--concurrency`-wide wave for those.

## Verification

- **685 tests pass** — also per test file standalone, in reverse module order, and with
  `GOODDATA_TOKEN`, `TAVERN_E2E_SKIP_TRACE_LINK`, `GD_EVAL_TIMERS` and `GD_EVAL_JUDGE_MODEL`
  exported. `ruff check`, `ruff format --check`, `uv run ty` and `uv lock --check` clean.
- **Pass rates unchanged against a live baseline:** 18/18 on `general_question` and 33/33 on the
  33-item `guardrail` dataset, on this branch and on the pre-change tree. `pass^K` differed on
  guardrail (32 vs 29); re-running the three differing items three times per tree gave master
  2/3, 1/3, 2/3 and this branch 2/3, 2/3, 2/3 — agent non-determinism on borderline refusal
  prompts, not a regression.
- **Trace lookups confirmed against production Langfuse:** zero `no trace found` across the
  runs above, slowest item 15.4s against a 120s budget.

Fixes found while reviewing this change — the judge contract, an interrupt that ran the whole
queue before stopping, the `sessionId` filter that never reached the server, and several tests
that certified nothing — are described in the commit message, each reproduced before it was
fixed.

## Impact on the gdc-nas nightly: none until the pin moves

`gdc-nas` installs from PyPI and pins `gooddata-eval == 1.73.1.dev3`, and `uv sync` runs without
`--upgrade`, so the nightly is unaffected until that lock is deliberately bumped. One thing was
fixed because of it: the tavern shims link traces **inline**, not batched, so they keep a **35s**
budget (identical to the old behaviour) while the batched CLI path gets **120s**. Land the lock
bump as its own gdc-nas PR and diff the pass rates before scheduling it.

## Known and deliberately deferred

- The link budget is per item, so at `--runs 3+` a slow first conversation shortens the later
  ones' retry ladders.
- `PARALLEL_SAFE_TEST_KINDS` classifies by evaluator; nothing restricts the agent's toolset, so
  it encodes an assumption about the fixtures rather than an enforced property.
- Only 2 of 8 agentic kinds record per-phase timings; the rest report `0.0` for
  agent/judge/simulated-user. `best_run_latency_s` is still `null` for agentic items.
- For multi-turn kinds the Langfuse cost comes from the single slowest trace in the session, so
  cost is under-counted.
- `_MAX_WORKERS = 16` for the drain pool ignores `--concurrency`; a 0% pass rate still exits 0;
  `WorkspaceModelController.restore(None)` no-ops; the four `GOODDATA_EVAL_CHAT_*` retry
  variables are undocumented.

## Summary by CodeRabbit

* **New Features**
  * Added configurable judge models and optional timing diagnostics.
  * Added safe parallel execution for supported evaluations.
  * Added support for passing user context through evaluations.
  * Added deferred Langfuse trace linking with improved interruption handling.
  * Added richer results, including run counts, pass-at-K, unanimous-pass status, and phase latency.

* **Bug Fixes**
  * Improved handling of incomplete judge responses and dataset test-kind detection.
  * Added warnings for local datasets used with Langfuse credentials.
  * Improved trace matching to prevent unrelated results from being selected.

* **Documentation**
  * Expanded CLI and JSON report documentation.

### DIFF
--- a/packages/gooddata-eval/pyproject.toml
+++ b/packages/gooddata-eval/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
-llm-judge = ["openai>=1.40,<2.0"]
+llm-judge = ["openai>=1.45,<2.0"]
 
 [project.scripts]
 gd-eval = "gooddata_eval.cli.main:main"
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/_langfuse.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/_langfuse.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 import base64
 import logging
 import os
+import threading
 import time
 import uuid
 from collections.abc import Iterator
@@ -15,7 +16,8 @@ from typing import Any
 
 import httpx
 
-from gooddata_eval.core.config import ReasoningEffort, normalize_reasoning_effort
+from gooddata_eval.core.agentic._trace_linker import link_cancel_event, linking_is_inline, warn_from_worker
+from gooddata_eval.core.config import ReasoningEffort, env_flag, normalize_reasoning_effort
 
 _log = logging.getLogger(__name__)
 
@@ -44,14 +46,37 @@ class _TraceAPI:
     def __init__(self, client: httpx.Client) -> None:
         self._client = client
 
-    def list(self, from_timestamp: Any, to_timestamp: Any, limit: int) -> _TraceListResult:
+    def list(
+        self, from_timestamp: Any, to_timestamp: Any, limit: int, session_id: str | None = None
+    ) -> _TraceListResult:
+        """List traces in a window, optionally narrowed to one session server-side.
+
+        ``session_id`` is what makes ``limit`` a non-issue. Without it the endpoint returns
+        every trace in the window newest-first and the caller filters locally, so an eval
+        workspace busy enough to put more than ``limit`` traces inside one item's window
+        pushes that item's OWN (oldest) trace off the page -- it then polls its whole retry
+        budget against a page that can never contain it, and the score orphans with only a
+        generic "no trace found" line to show for it. Concurrency makes that likelier by
+        overlapping every item's window. Named ``session_id`` because
+        ``_fetch_traces_for_session`` probes for exactly that parameter before it will stop
+        filtering locally; gen-ai sets sessionId = conversationId.
+        """
+
         def _ts(v: Any) -> str:
             return v.isoformat() if hasattr(v, "isoformat") else str(v)
 
-        resp = self._client.get(
-            "/api/public/traces",
-            params={"fromTimestamp": _ts(from_timestamp), "toTimestamp": _ts(to_timestamp), "limit": limit},
-        )
+        params: dict[str, Any] = {
+            "fromTimestamp": _ts(from_timestamp),
+            "toTimestamp": _ts(to_timestamp),
+            "limit": limit,
+        }
+        # `is not None`, not truthiness: _fetch_traces_for_session puts session_id into its
+        # kwargs unconditionally and then skips local filtering because it is there, so an
+        # empty id dropped here would return the whole padded window unfiltered -- and the
+        # max-latency pick would attach this item's scores to a stranger's trace.
+        if session_id is not None:
+            params["sessionId"] = session_id
+        resp = self._client.get("/api/public/traces", params=params)
         resp.raise_for_status()
         return _TraceListResult([_TraceObj(t) for t in resp.json().get("data", [])])
 
@@ -161,6 +186,15 @@ def make_langfuse_client() -> HttpxLangfuseClient:
     return HttpxLangfuseClient()
 
 
+def langfuse_credentials_present() -> bool:
+    """Whether the environment could build a Langfuse client.
+
+    Separate from ``try_make_langfuse_client`` so a caller can ask the question without
+    opening an httpx client it does not intend to use.
+    """
+    return bool(os.environ.get("LANGFUSE_PUBLIC_KEY")) and bool(os.environ.get("LANGFUSE_SECRET_KEY"))
+
+
 def try_make_langfuse_client() -> HttpxLangfuseClient | None:
     """Create Langfuse client from env vars; return None if credentials are missing."""
     try:
@@ -173,6 +207,28 @@ def try_make_langfuse_client() -> HttpxLangfuseClient | None:
 
 SKIP_ENV_VAR = "TAVERN_E2E_SKIP_TRACE_LINK"
 
+# Run names whose dataset-run assembly has already been reported as impossible. A 404 from
+# dataset-run-items means the dataset item id is not in Langfuse, which is a property of
+# the dataset and not of the attempt -- so it recurs identically for every item and every
+# run of that dataset, and reporting it per conversation buries the run's real output under
+# dozens of copies of the same HTTP error. Guarded by a lock because linking runs on the
+# drain pool's worker threads.
+_UNLINKABLE_RUNS: set[str] = set()
+_UNLINKABLE_RUNS_LOCK = threading.Lock()
+
+
+def _first_report_for_run(run_name: str) -> bool:
+    """True the first time this run is seen, False afterwards. Thread-safe."""
+    # A run is suffixed _run0.._runK per K, one per pass over the same dataset; the cause
+    # is shared across all of them, so collapse to the base name.
+    base = run_name.rsplit("_run", 1)[0]
+    with _UNLINKABLE_RUNS_LOCK:
+        if base in _UNLINKABLE_RUNS:
+            return False
+        _UNLINKABLE_RUNS.add(base)
+        return True
+
+
 _MAX_LATENCY_SEC = 60.0
 _MAX_COST_USD = 0.05
 _QUALITY_WEIGHT = 0.6
@@ -180,8 +236,22 @@ _SPEED_WEIGHT = 0.2
 _COST_WEIGHT = 0.2
 
 _INITIAL_DELAY = 0.5
-_MAX_ATTEMPTS = 8
 _BACKOFF = 1.6
+# Ceiling on any single backoff sleep, so a long budget is spent on many steady retries
+# rather than one enormous final wait.
+_MAX_DELAY = 15.0
+# Budget for one item's batched poll, which blocks nobody. Sized against Langfuse ingestion
+# lag on us.cloud, which runs from tens of seconds to several minutes. Shared across the
+# item's conversations, so the batch tail does not grow with --runs.
+_LINK_BUDGET_SEC = 120.0
+# Budget for a poll that is NOT batched: a direct library caller (the tavern e2e suite) gets
+# run_trace_link_inline, so the wait lands on the test that triggered it, under a step
+# timeout. Deliberately tighter than _LINK_BUDGET_SEC: no inline caller should pay for a
+# budget sized for the CLI's batched tail.
+_INLINE_LINK_BUDGET_SEC = 35.0
+# Sanity bound so the retry loop can never spin: the deadline is wall-clock and only the
+# sleeps advance it. Comfortably above the ~12 attempts the budget actually affords.
+_MAX_ATTEMPTS = 20
 _WINDOW_PADDING_SEC = 2
 _FETCH_LIMIT = 100
 
@@ -243,38 +313,100 @@ def _fetch_traces_for_session(
     return traces
 
 
+# Longest a running poll may stay asleep after cancellation is signalled. Without a bound,
+# a Ctrl-C mid-drain waits out the rest of that poll's backoff -- up to _MAX_DELAY per sleep
+# and _LINK_BUDGET_SEC overall -- because the interpreter joins executor workers at exit.
+_CANCEL_CHECK_SEC = 0.5
+
+
+def _wait_between_attempts(delay: float) -> bool:
+    """Wait ``delay`` before the next poll attempt. False means "stop polling".
+
+    Outside a batched drain nobody can cancel the wait -- an inline poll is charged to the
+    caller that asked for it -- so it stays a single sleep. Inside a drain the wait is served
+    in slices, so an interrupt is noticed within ``_CANCEL_CHECK_SEC`` rather than after the
+    whole backoff. Slicing keeps the total unchanged, so the retry ladder is unaffected.
+    """
+    cancel = link_cancel_event()
+    if cancel is None:
+        time.sleep(delay)
+        return True
+    remaining = delay
+    while remaining > 0:
+        if cancel.is_set():
+            return False
+        step = min(_CANCEL_CHECK_SEC, remaining)
+        time.sleep(step)
+        remaining -= step
+    return not cancel.is_set()
+
+
 def find_traces_per_conversation(
     langfuse: Any,
     conversation_ids: list[str],
     window_start: datetime,
+    window_end: datetime | None = None,
+    deadline: float | None = None,
 ) -> dict[str, Any]:
-    """Poll Langfuse until traces matching all conversation_ids are found or retries exhaust."""
-    if bool(os.environ.get(SKIP_ENV_VAR)):
+    """Poll Langfuse until traces matching all conversation_ids are found or retries exhaust.
+
+    ``window_end`` bounds the trace query and should be pinned by the caller to the moment
+    the conversations ended. It matters because this poll is normally deferred onto a
+    worker thread (see ``agentic/_trace_linker.py``): defaulting it to "now" would stretch
+    the window by however long the task waited in the queue, and since
+    ``_fetch_traces_for_session`` pages at ``_FETCH_LIMIT`` and filters by session locally,
+    a wide enough window can push the wanted trace off the page. Defaults to now only for
+    direct callers that poll immediately.
+    """
+    if env_flag(SKIP_ENV_VAR):
+        # Say so. Skipping returns all-None, which downstream renders as observe()'s
+        # generic "No trace found for dataset run ...; scores will be orphaned" -- the
+        # same message a real lookup failure produces. Left silent, an eval run looks
+        # like Langfuse is broken when trace linking was simply switched off.
+        warn_from_worker(
+            f"[langfuse] trace linking SKIPPED by {SKIP_ENV_VAR}: "
+            f"{len(conversation_ids)} conversation(s) will have orphaned scores. "
+            f"Unset it to link traces."
+        )
         return dict.fromkeys(conversation_ids)
 
     by_conv: dict[str, Any] = dict.fromkeys(conversation_ids)
-    window_end = datetime.now(timezone.utc)
+    window_end = window_end or datetime.now(timezone.utc)
     pad = timedelta(seconds=_WINDOW_PADDING_SEC)
 
+    # One budget for the whole item. Per conversation it would multiply by --runs, and the
+    # batch tail is meant to be bounded by the budget however many runs an item has.
+    budget = _INLINE_LINK_BUDGET_SEC if linking_is_inline() else _LINK_BUDGET_SEC
+    stop_at = deadline if deadline is not None else time.monotonic() + budget
+
     for cid in conversation_ids:
+        cancel = link_cancel_event()
+        if cancel is not None and cancel.is_set():
+            # The run is being interrupted; the remaining conversations are not worth a
+            # round trip, and their scores were never going to be written.
+            break
         delay = _INITIAL_DELAY
         found: list[Any] = []
-        for _ in range(_MAX_ATTEMPTS):
-            time.sleep(delay)
+        for _attempt in range(_MAX_ATTEMPTS):
+            # Attempt first, sleep only between attempts: a trace that is already ingested
+            # when we look must cost nothing, which is the common case once linking is
+            # batched to the end of the run.
             try:
                 found = _fetch_traces_for_session(langfuse, cid, window_start, window_end, pad)
             except Exception as exc:
                 _log.debug("Langfuse trace fetch failed for %s: %s", cid, exc)
-            if found:
+            if found or time.monotonic() + delay > stop_at:
                 break
-            delay *= _BACKOFF
+            if not _wait_between_attempts(delay):
+                break
+            delay = min(delay * _BACKOFF, _MAX_DELAY)
         if found:
             by_conv[cid] = max(found, key=lambda t: getattr(t, "latency", None) or 0.0)
         else:
             _log.warning(
                 "[langfuse] No trace found for conversation %s in window [%s, %s]", cid, window_start, window_end
             )
-            print(f"[langfuse] WARNING: no trace found for conversation {cid}", flush=True)
+            warn_from_worker(f"[langfuse] WARNING: no trace found for conversation {cid}")
 
     return by_conv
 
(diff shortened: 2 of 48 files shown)


packages/gooddata-eval/src/gooddata_eval/core/reporting/console.py
```diff
@@ -41,17 +50,24 @@ def render_console(report: EvalReport, *, console: Console | None = None) -> str
         latency = "-" if item.runs == 0 else f"{item.latency_s:.2f}s"
         avg = "-" if item.runs == 0 else f"{item.avg_latency_s:.2f}s"
         quality = "-" if item.skipped else f"{item.quality_score:.0%}"
-        table.add_row(item.id, item.test_kind, result, str(item.runs), latency, avg, quality, notes)
+        runs_col = str(_runs_total(item))
+        table.add_row(item.id, item.test_kind, result, runs_col, latency, avg, quality, notes)
```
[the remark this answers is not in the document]
tychtjan: Fixed, but on `ItemReport.avg_latency_s` rather than at the call site — the JSON report reads the same property, so fixing only `console.py` would have left `avg_latency_s` in the report still dividing by the requested K. It now divides by `runs_total`, so the Runs column and the average share one divisor. Covered by `test_avg_per_run_divides_by_the_runs_actually_taken`, verified to fail when reverted.

packages/gooddata-eval/src/gooddata_eval/core/dataset/langfuse_source.py
```diff
@@ -47,32 +49,57 @@ def _question_from_input(raw_input: Any) -> str:
...
+    rubric has a plain *string* expectedOutput, which gives the structure checks below
+    nothing to read -- metadata is the only place such a dataset can carry its own kind.
+    """
+    eo: dict[str, Any] | None = cast("dict[str, Any]", expected_output) if isinstance(expected_output, dict) else None
+    # An explicit declaration wins over structure, expected_output ahead of metadata.
+    # A blank declaration is not a declaration: "" would beat both the structural checks
+    # below and the CLI default, and the item would be skipped as an unsupported kind.
+    declared = _first_of(str, "test_kind", eo, metadata)
+    if declared and declared.strip():
+        return declared.strip()
```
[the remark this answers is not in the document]
tychtjan: Confirmed and fixed — and this was a regression introduced by the previous fix in this same function: `_first_of` returns the first *string* it finds, so a blank `expectedOutput.test_kind` shadowed a valid `metadata.test_kind`. Reproduced (`eo=blank, metadata=agentic_guardrail` resolved to `visualization`), now checked one source at a time. Covered by `test_a_blank_declaration_does_not_hide_a_real_one_behind_it`, verified to fail when reverted.

packages/gooddata-eval/src/gooddata_eval/core/agentic/_trace_linker.py
```diff
@@ -0,0 +1,266 @@
...
+        # mid-batch would run every QUEUED poll -- almost entirely time.sleep -- to
+        # completion first. abandon() cannot help here (the queue moved into `queue` above),
+        # so the cancellation has to happen on the pool itself.
+        pool = ThreadPoolExecutor(max_workers=min(len(queue), self._max_workers), thread_name_prefix="trace-link")
+        try:
+            for task, item_id in queue:
+                pool.submit(self._run, task, item_id)
+            pool.shutdown(wait=True)
+        except BaseException:
+            pool.shutdown(wait=False, cancel_futures=True)
```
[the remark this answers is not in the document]
tychtjan: Confirmed and fixed. Reproduced end to end: with a never-resolving fetch and four in-flight polls, the worker threads kept running for **110s** after the interrupt (`drain()` itself returned in 0.00s — the wait is the interpreter joining executor workers at exit). Now **0.50s**.

The linker publishes a `threading.Event` for the duration of one `drain()` and sets it *before* `shutdown(wait=False, cancel_futures=True)`; `find_traces_per_conversation` checks it between conversations and serves each backoff in `_CANCEL_CHECK_SEC` slices. The Event is created per drain and cleared in a `finally`, deliberately: a module-level one would let a Ctrl-C in one `--model` pass stop the next one’s polls before they started. The inline path keeps a single plain `time.sleep`, since there is nobody to cancel a poll charged to its own caller.

Two tests, both mutation-verified: `test_an_interrupt_signals_cancellation_before_it_shuts_the_pool_down` (asserts the signal/shutdown ordering) and `test_cancelling_mid_backoff_stops_the_poll_without_finishing_the_sleep` (cancels while the poll is already asleep, so it pins the sliced wait rather than the loop-top guard — my first attempt at this test passed even with the slicing removed).

packages/gooddata-eval/src/gooddata_eval/core/agentic/_langfuse.py
```diff
@@ -44,14 +46,33 @@ class _TraceAPI:
...
-            "/api/public/traces",
-            params={"fromTimestamp": _ts(from_timestamp), "toTimestamp": _ts(to_timestamp), "limit": limit},
-        )
+        params: dict[str, Any] = {
+            "fromTimestamp": _ts(from_timestamp),
+            "toTimestamp": _ts(to_timestamp),
+            "limit": limit,
+        }
+        if session_id:
+            params["sessionId"] = session_id
```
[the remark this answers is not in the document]
tychtjan: Confirmed and fixed. Reproduced against a mock transport: with `session_id=""`, `sessionId` was omitted from the query while `_fetch_traces_for_session` still skipped local filtering (it keys off the parameter being *present* in its kwargs, which it is), so the call returned two foreign traces — and `max(..., key=latency)` would have linked this item's scores onto the slower stranger.

Now an explicit `is not None` check, so an empty id is sent through and the server correctly matches nothing. Covered by `test_an_empty_conversation_id_still_sends_the_server_side_filter`, asserted at the HTTP layer (my first attempt stubbed `trace.list` directly and so could not see the defect at all, since it lives in the param building). Mutation-verified: reverting to the truthiness check fails the test.

packages/gooddata-eval/pyproject.toml
```diff
@@ -30,7 +30,9 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
-llm-judge = ["openai>=1.40,<2.0"]
+# >=1.45: max_completion_tokens (which _llm_judge now always sends) does not exist in
+# 1.40-1.44 and create() has no **kwargs there, so every judge call would TypeError.
```
hkad98: Nit: remove this description. It should be part of the commit message. Not here.

tychtjan: Removed.

packages/gooddata-eval/src/gooddata_eval/core/evaluators/summary.py
```diff
@@ -72,25 +72,49 @@ def evaluate(self, item: DatasetItem, chat_result: ChatResult) -> ItemEvaluation
...
 
         for i, criterion in enumerate(rubric):
-            ok, reason = self._positive_judge.score(item.question, criterion, actual)
-            detail[f"rubric_{i}"] = ok
-            detail[f"rubric_{i}_reason"] = reason
+            # Rubric criteria inform quality but never gate `passed`.
+            _grade(self._positive_judge, criterion, f"rubric_{i}")
 
         bool_checks = [v for v in detail.values() if isinstance(v, bool)]
+        if ungraded and not bool_checks:
```
hkad98: **An item can be reported PASS with zero mandatory criteria actually graded.**

`passed = passed and (ok is not False)` makes an ungraded criterion a no-op, and this guard is `if ungraded and not bool_checks` — but `bool_checks` collects *every* bool in `detail`, including the `rubric_*` entries, which by design never gate `passed`. So a single graded rubric criterion satisfies the guard while both gating criteria went ungraded.

Reproduced with `must_include: ["FACT A", "FACT B"]` both raising `JudgeResponseError` (a reasoning judge hitting `JUDGE_MAX_COMPLETION_TOKENS` twice) and one `rubric` criterion grading True:

```
passed = True   rank_key = (1, 1.0)
detail = {'actual_output': ..., 'rubric_0': True, 'ungraded_criteria': 2}
```

`quality_score` 100%, console shows a clean `PASS`, and not one criterion that gates the verdict was assessed. Worse than the pre-PR silent FAIL, because a false PASS doesn't get investigated.

Either key the guard on the gating criteria only:

```python
gating = len(must_include) + len(must_not_include)
graded_gating = sum(1 for k, v in detail.items() if isinstance(v, bool) and not k.startswith("rubric_"))
if gating and not graded_gating:
    raise JudgeResponseError(...)
```

or let an ungraded gating criterion disqualify the pass (`passed = passed and ok is True`). I'd lean to the second: "the judge could not tell us whether a mandatory fact is present" is not evidence that it is.

packages/gooddata-eval/src/gooddata_eval/core/evaluators/_llm_judge.py
```diff
@@ -40,27 +152,147 @@ def __init__(self, evaluation_steps: list[str], model: str = "gpt-4o"):
...
+    *exclude* it rather than count it, because a judge fault is a property of one request,
+    not of the agent's answer.
+    """
+
+    passed: bool
+    reasoning: str
+    error: str | None = None
+
+
+def score_run(judge: LLMJudge, *, input: str, expected_output: str, actual_output: str) -> JudgeVerdict:
```
hkad98, line 294: **`score_run` never reaches the two single-turn kinds, so they regress instead.**

`evaluators/general_question.py:25` and `evaluators/guardrail.py:42` still call `self._judge.score(...)` directly. Their agentic twins do use `score_run` (`agentic/general_question.py:115`, `agentic/guardrail.py:111`), and both non-agentic kinds are a live path — `general_question`/`guardrail` are absent from `AGENTIC_TEST_KINDS`, so they route through `run_items` at `cli/main.py:443`.

Reproduced with an item that passes run 1 and hits a judge fault on run 2 of 5:

```
pass_at_k = True | error = True | runs = 1
report.passed = 1   errored = 1   total = 1
```

Runs 3-5 are abandoned, and the item is counted in `EvalReport.passed` *and* rendered `ERROR` (`reporting/console.py:30` tests `error` before `pass_at_k`), so the JSON summary's `passed + failed + errored + skipped` stops reconciling with `total`.

This is new here rather than pre-existing: `master`'s `_llm_judge.py` has only `score`, returning `int(data.get("score", 0))`, and never raises. So the PR hands these two kinds a new item-level failure mode while giving the mitigation to the agentic path only — which is exactly what the docstring right here warns about ("discarded every run already graded ... reported an item whose pass@K was already satisfied as a failure").

`score_run` returns the shape both evaluators need. The only decision is what an ungraded run should mean for a single-criterion evaluator; simplest consistent answer is to raise only when *no* run graded, as `summary.py` intends. If it's left as-is, the PR body ("A judge fault on one run no longer errors the whole item") and the README's `--judge-model` section want a qualifying line.

packages/gooddata-eval/src/gooddata_eval/core/agentic/_langfuse.py
```diff
@@ -44,14 +46,37 @@ class _TraceAPI:
     def __init__(self, client: httpx.Client) -> None:
         self._client = client
 
-    def list(self, from_timestamp: Any, to_timestamp: Any, limit: int) -> _TraceListResult:
+    def list(
+        self, from_timestamp: Any, to_timestamp: Any, limit: int, session_id: str | None = None
```
hkad98, line 50: **Adding `session_id` here silently kills the local `sessionId` post-filter that was live on `master`.**

`_fetch_traces_for_session` probes `inspect.signature(langfuse.api.trace.list)` and only filters locally `if "session_id" not in kwargs`. On `master` this signature had no `session_id`, so the probe found nothing and the local filter **always** ran for the built-in httpx client. With the parameter declared, `kwargs["session_id"]` is now always set and the filter at line 307 is unreachable.

That matters when a server doesn't honour the param — the Langfuse API drops unknown query params rather than 400-ing. With a mock transport that sends `sessionId=mine` and ignores it:

```
traces returned: ['t-other', 't-mine']
picked: t-other          # max(found, key=latency)
```

The item's own trace is right there in the payload, and the filter that used to run would have picked it. Instead the scores and the dataset-run item attach to an unrelated conversation with no warning — strictly worse than the old "no trace found for conversation", which at least tells you the link is missing.

I realise the coupling is deliberate (this docstring and the `is not None` comment both spell it out), but the safety net cost 6 lines that already exist and `_TraceObj` already parses `sessionId`. Making it an unconditional post-check costs one deleted `if` and leaves the fast path intact:

```python
traces = response.data or []
# Post-check, not a fallback: the param is always sent now, but a server that ignores it
# would otherwise hand the max-latency pick a stranger's trace.
return [t for t in traces if _matches_session(t, session_id)]
```

packages/gooddata-eval/src/gooddata_eval/core/agentic/_trace_linker.py
```diff
@@ -0,0 +1,297 @@
...
+            _ACTIVE_CANCEL.set()
+            pool.shutdown(wait=False, cancel_futures=True)
+            raise
+        finally:
+            # Cleared only when nothing was cancelled. Leaving a SET event in place is
+            # deliberate: shutdown(wait=False) returns before the workers notice, and a
+            # worker that reaches its next wait after this line would otherwise read None
+            # and go back to an uninterruptible sleep for the rest of its budget. The next
+            # drain installs a fresh event, so a set one cannot leak into it.
+            if not _ACTIVE_CANCEL.is_set():
```
hkad98: **A cancelled drain leaves `_ACTIVE_CANCEL` set forever, which disables later *inline* linking.**

The reasoning for leaving a set event in place holds for the drain itself (workers still in flight need to see it, and the next drain installs a fresh one). What it misses is the inline path: `run_trace_link_inline` neither installs nor clears the global, so `find_traces_per_conversation` reads a set event at the top of its `for cid` loop and breaks before a single fetch.

```
after interrupted drain: link_cancel_event() = <Event: set>
a later INLINE link sees cancel set = [True]
find_traces_per_conversation -> {'c1': None, 'c2': None}   # no fetch attempted
```

In the shipped CLI this is unreachable — `drain()` is called only at `cli/agentic_runner.py:440` and the interrupt re-raises, so the process exits. It needs a caller that catches `KeyboardInterrupt` and carries on.

There is one such process today: the test suite. `tests/test_trace_linker.py:285` leaks the set event, and `test_no_cancellation_signal_exists_outside_a_drain` (line 376, whose first assert is `link_cancel_event() is None`) passes only because the normal drain at line 335 happens to run in between and resets it. Shuffle test order and that assert fails for a reason unrelated to what it tests. Worth closing for the suite's sake if not the runtime's — scope the event to the linker instance, or clear a stale one at the top of `run_trace_link_inline`.

packages/gooddata-eval/src/gooddata_eval/core/evaluators/_llm_judge.py
```diff
@@ -1,12 +1,124 @@
...
+    if isinstance(body, dict):
+        error = body.get("error")
+        error = error if isinstance(error, dict) else body
+        if error.get("param") == "temperature":
+            return True
+        message = error.get("message")
+        return "temperature" in message.lower() if isinstance(message, str) else False
+    if getattr(exc, "param", None) == "temperature":
+        return True
+    return "temperature" in str(exc if body is None else body).lower()
```
hkad98: **This falls into the trap the docstring above warns about.**

The docstring says to read this off the provider's structured error and "never off `str(exc)`", because gateways echo the request back inside the body — and then this line stringifies `body` whenever it is present and *not* a dict. The openai SDK sets `body` to `response.text` when the error body doesn't parse as JSON, so the branch is reachable:

```python
class Exc(Exception):
    body = '{"error":{"message":"model not found"},"request":{"temperature": 0}}'

_rejects_temperature(Exc("bad request"))  # -> True
```

Cost is bounded — a misleading `verdicts are NOT deterministic` warning, `_supports_temperature` latched off for the rest of the item, one wasted request, and the real error surfacing only from the retry — so this is the smallest of the five. Restricting the fallback to prose matches what the docstring already argues for:

```python
return "temperature" in str(exc).lower()
```

hkad98 (commented):
Reviewed the full diff at `d4bd0e75`; suite is green (685 with the `llm-judge` extra). The trace-linking refactor itself checks out — all eight `evaluate_agentic_*` take `submit_trace_link` and populate `runs_passed`/`runs_effective`, `dataset_item_id` lines up with `ItemReport.id` for the `linker.durations` lookup, no `_langfuse` -> `_trace_linker` import cycle, and the drain is always ordered before `langfuse.close()` and before any report renders.

Five findings inline, each with a repro. Two I'd want fixed before merge, both in the judge-validation half and both the same shape as the bug this PR exists to kill — a judge fault becoming a verdict nobody assessed:

1. `dashboard_summary` can report **PASS with zero mandatory criteria graded** (`evaluators/summary.py:110`).
2. The `score_run` mitigation **doesn't reach the non-agentic `general_question`/`guardrail`**, which the PR body and README both say it does — and since `JudgeResponseError` is new here, those two kinds come out of this PR with a *new* item-level failure mode.

One more that is a behaviour regression rather than a nit: adding `session_id` to `_TraceAPI.list` silently killed the local `sessionId` post-filter that was live on `master`.

Not raised, since the PR body already discloses them: the shared per-item link budget, and `_MAX_WORKERS = 16` ignoring `--concurrency`. Two smaller notes: `WORKSPACE_MUTATING_TEST_KINDS` (`cli/agentic_runner.py:80`) is read by tests and named in the runner docstring but no production code branches on it (`runs_in_parallel` goes through `PARALLEL_SAFE_TEST_KINDS`); and `JUDGE_MAX_COMPLETION_TOKENS` is sent unconditionally with no graceful degrade, unlike `temperature`, so an endpoint that accepts only `max_tokens` (older Azure `api-version`s) would hard-fail every judge call — `--judge-model` invites non-OpenAI targets, though I couldn't confirm a concrete one in use here.

packages/gooddata-eval/src/gooddata_eval/core/evaluators/summary.py
tychtjan: Fixed in 7f04d926, the second option. An ungraded gating criterion now disqualifies the pass (`ok is True`), and the no-verdict guard keys on gating criteria, so a graded rubric line cannot certify a pass. Your repro is `test_a_graded_rubric_cannot_carry_a_pass_when_every_gating_criterion_went_ungraded`. The evaluator also returns the ungraded run instead of raising, so one all-ungraded run out of K no longer errors the item; the runner errors only when no run was graded.

packages/gooddata-eval/src/gooddata_eval/core/evaluators/_llm_judge.py
tychtjan: Fixed in 7f04d926. Both evaluators go through `score_run` and return an ungraded run via `ItemEvaluation.error`. The runner counts those in a new `runs_ungraded` (JSON + console Notes), keeps evaluating the remaining runs, and errors the item only when none was graded. Your repro is `test_one_ungraded_run_does_not_error_an_item_whose_other_run_passed`: run 1 pass, run 2 fault, both runs taken, PASS with 1/2 passed, 1 ungraded, `passed + failed + errored + skipped == total` again.

packages/gooddata-eval/src/gooddata_eval/core/agentic/_langfuse.py
tychtjan: Fixed in 7f04d926, as you suggested: the local filter is an unconditional post-check now, the server-side parameter is still sent. `test_a_server_that_ignores_the_session_filter_cannot_hand_over_a_foreign_trace` replays your mock (server returns t-other + t-mine) and asserts only t-mine survives.

packages/gooddata-eval/src/gooddata_eval/core/agentic/_trace_linker.py
tychtjan: Fixed in 7f04d926. The signal is a ContextVar set per task on the worker thread (`BackgroundTraceLinker._run`), not a module global: late workers keep seeing the set event, the calling thread never does, so neither a later inline link nor the next drain can inherit it, and `link_cancel_event() is None` on the main thread regardless of test order. `test_an_interrupted_drain_does_not_cancel_a_later_inline_link` covers your scenario.

packages/gooddata-eval/src/gooddata_eval/core/evaluators/_llm_judge.py
tychtjan: Fixed in 7f04d926, one step stricter than `str(exc)`: when the body is not JSON the openai SDK sets the exception message to that same raw text (`_make_status_error_from_response`), so a substring check on the message hits the same echoed request. A non-dict body is now not read at all; the message fallback stays only for exceptions with no body. Test: `test_an_unparseable_error_body_is_not_read_as_a_temperature_rejection`.

packages/gooddata-eval/src/gooddata_eval/core/agentic/_trace_linker.py
```diff
@@ -0,0 +1,297 @@
...
+                _ACTIVE_CANCEL = None
```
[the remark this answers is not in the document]
tychtjan: Addressed in 7f04d926 by scoping the event to the worker thread (ContextVar set per task) rather than special-casing the inline path; `run_trace_link_inline` never sees a drain's event at all.

packages/gooddata-eval/src/gooddata_eval/core/reporting/json_report.py
```diff
@@ -33,9 +40,23 @@ def _build_run_dict(report: EvalReport) -> dict:
                 "runs": item.runs,
```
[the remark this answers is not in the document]
tychtjan: Addressed in 7f04d926: JSON `runs` and `EvalReport.total_runs` use `runs_total`. `test_json_runs_and_the_aggregate_average_use_the_runs_actually_taken`.

hkad98 (approved):

PR #1784, opened 2026-09-07
fix(gooddata-eval): assert alert group-by attributes

tychtjan:
## Summary

An alert can be narrowed two ways: a date entry in `filters`, or a date group-by in
`attributes`. Only the first was checked. The second is worse — a group-by makes the alert fire
per period value instead of on the latest one, which is a different alert from the one the
fixture describes. An alert with `filters: []` and `attributes: [order_date.month]` therefore
scored 7/7 clean, and a recorded GPT-5.2 run carried exactly that invented month group-by and
passed.

**This activates coverage the dataset already carries**, it is not a forward-looking change.
Item `e2a1e22a-2020-4a24-b3fe-2350315f7a73` ("Alert me every time returns for any individual
product brand go above 30 in the last day") stores
`"Attributes": [{"using": "label/product_brand"}]`, and that key was dropped on the floor. It is
the regression gate for this change, not just the new unit tests. QA's rewritten `c2f8b8ed`
carrying `"Attributes": []` is the second beneficiary.

For the GDAI-2175 fix itself the eval could not serve as a before/after metric at all — it
reported 9/9 both before and after. That blind spot is what this closes.

### Why the two sides cannot be deep-compared

This is the trap the first attempt fell into. The expectation and the actual use different
vocabularies for the same grouping:

| side | shape |
|---|---|
| fixture `expected_output["Attributes"]` | AAC tool-input: `{"using": "label/x"}` |
| `create_metric_alert` argument | resolved AFM: `{"localIdentifier": "a0", "label": {"identifier": {"id": "x", "type": "label"}}}` |

The conversion happens in gdc-nas `prepare_metric_alert_proposal` (`_resolve_afm_slicing` →
`build_afm_execution_payload_from_query`), whose result is exposed as `afm_attributes`,
documented *"forward verbatim as create_metric_alert 'attributes'"*.

So `_attribute_label_ids` canonicalises both sides to a bare label id and the comparison is a
sorted multiset — order-insensitive, and the converter-assigned `localIdentifier` stops
mattering without a strip step. Identity is all that is compared; `showAllValues` is the agent's
to choose.

**The asymmetry is specific to `attributes`.** `Filters` in the same fixture is already
AFM-shaped (`relativeDateFilter`), which is why `_check_filters` gets away with a raw
`_deep_subset` and is left untouched here.

An unrecognised entry shape, or a URI prefix other than `label/`, raises — `label/x` and
`attribute/x` are different objects, and a new spelling must fail loudly rather than compare
unequal and read as the agent being wrong. The expectation side is validated in
`_normalize_expected_attributes`, so a malformed fixture fails before the run spends an API
call; either way `cli/agentic_runner.py` contains it to that item's own row.

### Decisions taken deliberately

- **`generate_simulated_alert_response` is unchanged.** An `Attributes: []` expectation needs no
  support: the simulated user does not invent a grouping and the check verifies it did not. A
  non-empty expectation does have a gap, but rule 3's "check ALL of these" list would have to
  grow too for a symmetric rule to be honest, perturbing all 18 items to serve the one that
  already requests its grouping in its own question. So a non-empty `Attributes` expectation
  requires the item's question to ask for that grouping, documented on the normalizer.
- **`core/evaluators/alert_skill.py` is untouched** — separate single-turn path, own tests, own
  ticket if parity is wanted.
- **The other half of GDAI-2175** — whether the agent *asked* before proposing — is not
  observable from final tool arguments and stays with `verify_alert_asks_for_date` in gdc-nas.

### Heads-up for dashboards

`quality_score` for alert items moves from /7 to /8, since both Langfuse sinks derive it as the
fraction of true booleans in the detail dict. Scores are not comparable across this commit, and
`attributes_correct` is a new Langfuse score name.

## Test plan

- [x] The comparison was verified red first: the previous `_deep_subset` returns `False` for both
      the product-brand and the `customer_created_date.month` case.
- [x] 19 tests in the attributes block, expectation side AAC-shaped against AFM-shaped actuals
      throughout — a test pairing AFM against AFM would prove nothing, which is how this slipped
      through once. Covers item `e2a1e22a`'s literal stored value as the regression gate, the
      granularity suffix on a date group-by, both spellings against one actual, the bare-string
      and `{"identifier": {"id": ...}}` forms, order-insensitivity, `showAllValues` ignored, and
      raises on each side.
- [x] `uv run --no-sync pytest tests/test_agentic_alert_skill.py tests/test_models.py tests/test_alert_skill_evaluator.py -q` — 83 passed.
- [x] `make format-fix lint-fix type-check test` — 732 passed on py310–py314, lint and types clean.
- [ ] Follow-up, not in this PR: cut a gooddata-eval dev release, then pin
      `tests/tavern-e2e/pyproject.toml` in gdc-nas to that exact version so QA's rewritten
      `c2f8b8ed` can carry a meaningful `"Attributes": []`.

risk: low

## Summary by CodeRabbit

- **New Features**
  - Alert evaluations now support expected group-by attributes.
  - Attribute matching recognizes equivalent labels regardless of input format or order.
  - Evaluation results include attribute correctness in pass/fail scoring and details.
  - Supports explicit no-grouping expectations and validation of malformed attribute inputs.

- **Tests**
  - Added comprehensive coverage for attribute normalization, canonicalization, ordering, missing values, prose, and invalid inputs.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/alert_skill.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/alert_skill.py
@@ -127,6 +127,78 @@ def _check_filters(expected: CatalogMetricAlert, actual_args: dict) -> bool:
     return _deep_subset(exp_filters, act_filters)
 
 
+def _attribute_label_ids(items: list, *, side: str) -> list[str]:
+    """Canonicalise group-by entries to bare label ids, whatever spelling they arrive in.
+
+    The two sides of the comparison speak different vocabularies for the same grouping.
+    Fixtures author the AAC tool-input form, ``{"using": "label/x"}``; ``create_metric_alert``
+    receives the resolved AFM form, ``{"localIdentifier": "a0", "label": {"identifier":
+    {"id": "x", "type": "label"}}}``, forwarded verbatim from ``prepare_metric_alert_proposal``.
+    Identity is therefore the only thing they can be compared on.
+
+    A shape not listed here, or a URI prefix other than ``label/``, raises: ``label/x`` and
+    ``attribute/x`` are different objects, and an unknown spelling must fail loudly rather
+    than quietly compare unequal.
+    """
+    if not isinstance(items, list):
+        raise ValueError(f"Unrecognised {side} group-by attributes, expected a list: {items!r}")
+    ids: list[str] = []
+    for item in items:
+        raw: object = None
+        if isinstance(item, str):
+            raw = item
+        elif isinstance(item, dict):
+            label = item.get("label")
+            identifier = item.get("identifier")
+            if isinstance(item.get("using"), str):
+                raw = item["using"]
+            elif isinstance(label, dict) and isinstance(label.get("identifier"), dict):
+                raw = label["identifier"].get("id")
+            elif isinstance(identifier, dict):
+                raw = identifier.get("id")
+        if not isinstance(raw, str) or not raw:
+            raise ValueError(f"Unrecognised {side} group-by attribute entry: {item!r}")
+        prefix, slash, rest = raw.partition("/")
+        if not slash:
+            ids.append(raw)
+        elif prefix == "label" and rest:
+            ids.append(rest)
+        else:
+            raise ValueError(f"Unrecognised {side} group-by attribute reference: {raw!r}")
+    return ids
+
+
+def _check_attributes(expected: CatalogMetricAlert, actual_args: dict) -> bool:
+    """Compare group-by identity only.
+
+    Per-entry properties — ``showAllValues``, the converter-assigned ``localIdentifier`` —
+    are deliberately not asserted, and the comparison is a multiset so entry order does not
+    matter.
+    """
+    exp_attributes = expected.attributes
+    if exp_attributes is None:
+        return True
+    act_attributes = actual_args.get("attributes")
+    if act_attributes is None:
+        # Arguments are raw `json.loads` output, where an unset nullable argument arrives as
+        # null rather than absent. Both spellings of "no grouping" have to land on [], which
+        # is why this is not `actual_args.get("attributes", [])`.
+        act_attributes = []
+    elif not isinstance(act_attributes, list):
+        # An argument that is not a list of groupings is the agent answering wrongly, so it
+        # scores False. Raising instead would make the runner record an ERROR, and errored
+        # items are excluded from the failure count — a malformed answer must not rank above
+        # a merely wrong one. An unreadable *entry* still raises, in `_attribute_label_ids`:
+        # entries are typed at the tool boundary, so the plausible cause there is the wire
+        # format moving, which has to be unmissable.
+        return False
+    if not exp_attributes:
+        return not act_attributes
+    exp_ids = sorted(_attribute_label_ids(exp_attributes, side="expected"))
+    act_ids = sorted(_attribute_label_ids(act_attributes, side="actual"))
+    return exp_ids == act_ids
+
+
 def _check_metric(expected: CatalogMetricAlert, actual_args: dict) -> bool:
     if not expected.metric_id:
         return True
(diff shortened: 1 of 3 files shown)


packages/gooddata-eval/src/gooddata_eval/core/agentic/alert_skill.py
```diff
@@ -127,6 +127,63 @@ def _check_filters(expected: CatalogMetricAlert, actual_args: dict) -> bool:
...
+
+def _check_attributes(expected: CatalogMetricAlert, actual_args: dict) -> bool:
+    """Compare group-by identity only.
+
+    Per-entry properties — ``showAllValues``, the converter-assigned ``localIdentifier`` —
+    are deliberately not asserted, and the comparison is a multiset so entry order does not
+    matter.
+    """
+    exp_attributes = expected.attributes
+    act_attributes = actual_args.get("attributes") or []
```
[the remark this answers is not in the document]
tychtjan: Valid, and wider than described — fixed in a182e2c8, though not with the proposed patch.

Two defects, not one. Besides the falsy case you found, `_attribute_label_ids` never checked that its input was a list, so a *truthy* non-list was iterated: a dict yields its keys, a string yields its characters. That gave a false pass — an actual of `"x"` scored correct against an expectation of `[{"using": "label/x"}]`.

On the remedy: raising here would misclassify the event. `json_report.py:22-25` draws an explicit line — *"a judge fault is an error, not K failures"* — and `"failed"` excludes items with `error is not None`, while `console.py:39` renders ERROR before it tests `pass_at_k`. So an exception means the harness could not judge, not that the subject was wrong. Raising on the agent's own output would drop a malformed answer out of the failure count entirely, ranking it above a merely wrong one.

But dropping the raise altogether loses something real: if the AFM encoding moves, a silent `False` across every item reads as an agent regression. So the split is structural rather than by side:

- **not a list of groupings at all** (`{}`, `""`, `0`, `"x"`) → `False`. The argument is wrong; that is a FAIL and must stay in the count.
- **a list holding an unreadable entry** → still raises. Entries are typed `AttributeItem` at the tool boundary, so the plausible cause is the wire format moving, which has to be unmissable.

The expectation side needed no equivalent guard: `_normalize_expected_output` already rejects a non-list before the run starts, and `from_dict` — the other way into the field — has no callers in the repo.

packages/gooddata-eval/src/gooddata_eval/core/agentic/alert_skill.py
```diff
@@ -127,6 +127,76 @@ def _check_filters(expected: CatalogMetricAlert, actual_args: dict) -> bool:
...
+    are deliberately not asserted, and the comparison is a multiset so entry order does not
+    matter.
+    """
+    exp_attributes = expected.attributes
+    if exp_attributes is None:
+        return True
+    # An unset argument serialises as null, so both spellings of "no grouping" land on [].
+    act_attributes = actual_args.get("attributes")
+    if act_attributes is None:
+        act_attributes = []
```
hkad98, line 186: Nit: 
```python
act_attributes = actual_args.get("attributes", [])
```

tychtjan: Can't take this one — `get(k, default)` only substitutes the default when the key is *missing*, and here it isn't.

An agent that adds no grouping doesn't omit the parameter: `create_metric_alert` serialises it as `"attributes": null`, so after `json.loads` the key is present holding `None`. The default is skipped, `None` reaches the `isinstance` check, and the result is `False` — a correct answer scored as a failure, on exactly the `Attributes: []` fixture this PR adds.

```python
{}.get("x", [])            # -> []    key missing -> default used
{"x": None}.get("x", [])   # -> None  key present -> default ignored
```

Same thing already bit us on `trigger`, and the comment at line 107 records it: *".get(k, default) only returns the default when the key is ABSENT, but create_metric_alert serialises unset params as `trigger: null`"*.

`or []` isn't available either — that was the earlier finding on this line: `{}` and `""` would collapse to `[]`, so a malformed argument would score as correct. One trap on each side, so both branches carry weight: `is None` for "nothing was passed", `isinstance` for "passed something that isn't a list".

Kept the code as is and added a comment naming the rejected form so it doesn't get re-proposed.

hkad98 (approved):

PR #1788, opened 2026-09-09
fix(gooddata-eval): three false-negative sources in eval scoring

myhoai:
QA-29230: capture every multipart response part, not just text/visualization/ alertProposal, and render them into what the judge and the simulated user see. QA-29226: normalize whitespace around MAQL punctuation before comparing, in one shared module the agentic and non-agentic metric comparators both use. QA-29225: read the fixture's anomaly granularity and stop telling the simulated user to refuse one, which deadlocked ANOMALY alert items.

## Summary by CodeRabbit

- **New Features**
  - Alerts now support hourly, daily, weekly, monthly, quarterly, and yearly anomaly-detection intervals.
  - Responses can include readable search results and other supported content, even without direct text.
  - Chat results now capture search results and additional non-text content.

- **Bug Fixes**
  - Improved fallback rendering for empty responses across conversations, metrics, KDA, visualizations, and guardrails.
  - Metric comparisons now better tolerate equivalent MAQL formatting, whitespace, and case variations.
  - Alert evaluations now validate anomaly-detection granularity more accurately.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/_catalog.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/_catalog.py
@@ -2,6 +2,41 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
+from enum import Enum
+
+
+class AnomalyDetectionGranularity(str, Enum):
+    """Detection intervals an anomaly alert accepts.
+
+    Mirrors gen-ai's enum of the same name; `StrEnum` is unavailable on the 3.10 floor, so
+    the `str` mixin carries the comparison against the raw tool argument.
+    """
+
+    HOUR = "HOUR"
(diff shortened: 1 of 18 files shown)


packages/gooddata-eval/src/gooddata_eval/core/agentic/_catalog.py
```diff
@@ -30,6 +30,8 @@ class CatalogMetricAlert:
     """Attribute filters applied to the alert condition."""
     attributes: list | None = None
     """Expected group-by attributes; ``None`` means the fixture states no expectation."""
+    granularity: str | None = None
```
hkad98: Nit: this granularity can be Enum – we know the values that can enter the granularity.

hkad98 (approved):

PR #1797, opened 2026-09-09
feat(gooddata-eval): evaluate the dashboard summary users actually get

Tomkess:
## The problem

The `dashboard_summary` test kind drives `POST /api/v1/ai/workspaces/{ws}/summary`. **Nothing in the product reaches that endpoint.**

Clicking a dashboard's **Summarize** menu item drops the user into the assistant with "Summarize this dashboard" pre-filled, so real summaries are produced by the conversational `dashboard_summary` skill. The endpoint is a separate capability for API/embedding consumers, gated by its own flag `ENABLE_GEN_AI_HEADLESS_SUMMARY` and enforced in `afm-exec-api` before gen-ai is reached.

Consequence: every `dashboard_summary` result ever recorded — 36 of them, across 12 consecutive days — is a `400`, and not one summary has ever been graded. The two paths share an idea and nothing else: different flag, different prompt, different tools, different failure modes.

This adds `agentic_dashboard_summary` for the path users take. It does not remove the existing kind — if someone ships an embedded summarize button, that coverage is still worth having.

## Making the skill engage

More than a dashboard id is needed, and the failure is silent. `dashboard_summary_skill` reads `userContext.view.dashboard` and collects only widgets carrying a `result_id`:

```python
for widget in user_context.view.dashboard.widgets:
    if widget.insight is not None:
        if widget.insight.result_id:   # no result_id -> dropped from the summarize scope
```

Probed live against `micdiagnose-dev`:

| Sent | Agent's reply |
|---|---|
| `dashboard.id` only | *"I can't summarize this dashboard… no dashboard charts were provided"* |
| id + widget descriptors, no `resultId` | *"these visualizations need to be reloaded"* |
| id + descriptors **with** `resultId` | a real summary, with real figures |

In the browser those ids exist because the client already rendered the widgets. The runner reproduces that deliberately: walk the layout, execute each insight via `sdk.compute.for_exec_def(...).result_id`, send what execution produced. Verified end to end against a 25-widget dashboard, all criteria graded.

## Scoring is unchanged

`DashboardSummaryEvaluator` grades free text against `must_include` / `must_not_include` / `rubric` and does not care which transport produced the text. The 32 existing fixtures port over by moving the dashboard id to the field this kind reads.

## Shape notes

- **Widgets are executed once and reused across K runs.** Result ids name cached executions, so re-running them per K would multiply the item's dominant cost without changing what the assistant sees.
- **Partial coverage is reported, not hidden.** A widget that fails to execute is left out of the context (the skill would drop it regardless) but still counted, and `detail` carries `widgets_executed` / `widgets_total`. A rubric naming a widget that never ran fails for a reason that is not the agent's, and that must be visible without opening the trace.
- **`max_widgets`** bounds the executions for wide dashboards — 29 widgets is 29 executions per item.
- **The dashboard entity is read over plain HTTP.** `entities_api` validates the entity's timestamp fields with a regex and raises `TypeError: expected string or bytes-like object, got 'datetime.datetime'`, so the typed accessor cannot read a dashboard at all today. Noted in the code; worth a separate fix in the generated client.
- **A `ChatError` is recorded on the run rather than raised**, so it cannot discard the K-runs already completed — same contract as #1789.

## Tests

15 new unit tests: layout walking (nested tabs/sections, rich-text skipping, dedupe), context building (result ids, failed-widget exclusion, `max_widgets`), run behaviour (coverage reporting, execute-once-across-K, chat-error recording and K-run preservation), and the `evaluate_*` wrapper (detail contents, raise-with-detail).

The two staleness guards — `_ALL_AGENTIC_KIND_CASES` and `_EVALUATE_FUNCS` — both caught the new kind as designed and are updated.

795 passed, lint and format clean.

## Follow-up, not in this PR

The eval repo's 32 fixtures need their dashboard id moved and `agentic_dashboard_summary` enabling. Two of the eight English fixtures also carry dashboard ids that no longer exist in the workspace — including the only one that ever ran — so those need re-pinning regardless of which kind they feed.

## Summary by CodeRabbit

- **New Features**
  - Added agentic dashboard-summary evaluations for assessing generated summaries against expected content.
  - Evaluations now gather context from dashboard visualizations, including nested layouts and widget results.
  - Added repeated-run scoring, widget coverage, timing details, and graceful reporting of visualization or chat failures.
  - Added support for configuring dashboard-summary evaluations through the evaluation runner.

- **Tests**
  - Added comprehensive coverage for dashboard traversal, context generation, execution reuse, error handling, scoring, and reporting.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
+++ b/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
@@ -12,6 +12,7 @@ from gooddata_eval.core.agentic._langfuse import make_langfuse_client
 from gooddata_eval.core.agentic._trace_linker import BackgroundTraceLinker, SubmitTraceLink, run_trace_link_inline
 from gooddata_eval.core.agentic.alert_skill import evaluate_agentic_alert_skill
 from gooddata_eval.core.agentic.conversation import ConversationFixture, evaluate_agentic_conversation
+from gooddata_eval.core.agentic.dashboard_summary import evaluate_agentic_dashboard_summary
 from gooddata_eval.core.agentic.general_question import evaluate_agentic_general_question
 from gooddata_eval.core.agentic.guardrail import evaluate_agentic_guardrail
 from gooddata_eval.core.agentic.kda_skill import evaluate_agentic_kda_skill
@@ -44,6 +45,7 @@ AGENTIC_TEST_KINDS = frozenset(
         "agentic_guardrail",
         "agentic_conversation",
         "agentic_kda_skill",
+        "agentic_dashboard_summary",
     }
 )
 
@@ -224,6 +226,23 @@ def _dispatch_agentic(
             workspace_id=workspace_id,
             question=item.question,
             expected_output=eo if isinstance(eo, dict) else {},
+            k=k,
+            agent_id=agent_id,
+            **lf_kw,
+        )
+    elif kind == "agentic_dashboard_summary":
+        summary_input = item.summary_input
+        if summary_input is None:
+            raise ValueError(f"agentic_dashboard_summary item '{item.id}' is missing required 'summary_input'.")
+        return evaluate_agentic_dashboard_summary(
+            host=host,
+            token=token,
+            workspace_id=workspace_id,
+            dashboard_id=summary_input.dashboard_id,
+            expected_output=eo,
+            # The fixture's own wording is the prompt under test -- a localized fixture is
+            # only meaningful if its own phrasing is what reaches the agent.
+            question=item.question,
             k=k,
             agent_id=agent_id,
             **lf_kw,
--- a/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
+++ b/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
@@ -243,6 +243,10 @@ def _dispatch_agentic(
             # The fixture's own wording is the prompt under test -- a localized fixture is
             # only meaningful if its own phrasing is what reaches the agent.
             question=item.question,
+            # Each widget costs an execution, so a fixture asserting on a handful of charts
+            # can name them instead of paying for the whole dashboard. Same field and meaning
+            # the headless /summary endpoint gives it; None summarizes everything.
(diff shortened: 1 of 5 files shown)


Tomkess:
All four addressed in `a86cd427`. Three were behavioural and all three were fair hits.

**`create_conversation` outside the handler.** Correct, and it's the same bug #1789 just fixed elsewhere — I guarded `send_message` and left the line directly above it bare, so a transient failure on run 2 of 3 discarded run 1. Recorded as a failed run now, with the empty `conversation_id` naturally excluded from trace scoring as you noted.

**`max_widgets` unreachable from a fixture.** Also correct — I built the cap, documented it as the cost control, then never wired it. Rather than forward the numeric cap, fixtures now bound cost by naming the charts they assert on via `summary_input.visualizations`: same field and meaning the headless `/summary` endpoint gives it, and it stays correct as a dashboard grows, which a positional cap does not. Names match against either the visualization id or the widget id, since a fixture author reading the AAC sees both. `max_widgets` stays as a blunt cap for direct callers and the docstring now says so.

**Private `gooddata_sdk.table` helpers.** Fair. `_execute_widget` now checks all three symbols up front and raises naming the dependency, so an SDK bump reads as a dependency problem instead of an `AttributeError` on the first widget of a run. I kept the private call rather than switching to the public API, and recorded why in the docstring: `sdk.tables.for_visualization` reads the whole result into an `ExecutionTable` and returns that, which discards the result id this needs and pays for every row to do it. Two tests pin it — one that the guard fires when the symbols vanish, one that it passes against the installed SDK so it can't become a permanent tripwire.

**The `_DEFAULT_PROMPT` comment.** Right, and the contradiction was within one commit: the comment said the prompt is kept verbatim while the dispatch passed `item.question` with the opposite rationale beside it. The dispatch is correct — a localized fixture is only meaningful if its own wording reaches the agent — so the comment now describes the constant as the default for direct callers.

Five tests added. The three behavioural ones were each verified to fail against the previous version before being kept. 800 passed, lint and format clean.

On the merge-risk summary: all three concerns it named are now closed. The remaining Docstring Coverage warning is mostly test functions, which carry their intent in their names and docstrings where the reasoning is non-obvious.

PR #1798, opened 2026-09-10
feat(gooddata-eval): add the agentic forecasting evaluator

Tomkess:
First of three evaluators for skills that ship in the product but have no eval coverage: forecasting, what-if analysis, anomaly detection. Each is a separate PR; this one is forecasting.

## Why now

Probed live against `micdiagnose-dev`: the skill **is** enabled and reachable — the agent answers a forecast question by activating `set_skills(["search", "forecasting", "visualization"])`. Nothing evaluates it.

## This one can check correctness, not just completion

`kda_skill` is deliberately scoped to "the process ran to completion". Forecasting does not have to be, because the skill refuses to execute unless the visualization carries an AAC forecast config — and that config is exactly where the user's request lands:

| Config field | Carries |
|---|---|
| `forecast_enabled` | must be `true`, or `execute_forecast` returns an error |
| `forecast_period` | the horizon — "next 3 months" is `3` |
| `forecast_confidence` | confidence level |
| `forecast_seasonal` | whether seasonality is modelled |

So "did it forecast the right horizon" is a number in the tool call the agent made. No judge, no paraphrase tolerance. The measure forecast is checked the same way, off the visualization's own fields.

A fixture pins whatever it cares about:

```json
{"metric": "metric/spend", "forecast_period": 3}
```

## Two details worth review

**An unstated expectation passes** rather than fails — but `detail["asserted"]` records which checks the fixture actually pinned. Without that, a run that verified nothing is indistinguishable in the report from one where everything matched.

**`forecast_enabled` must be explicitly `true`.** The tool treats unset and `false` the same way, so the check does too — an agent that builds the right chart but never enables forecasting has not done the job.

## The loop

Follows `kda_skill`. The agent routinely asks which measure to forecast before building anything — observed live: *"your data has two different Spend metrics that could mean different things."* A simulated user answers from the fixture's own hints, and only hints the fixture supplies reach the prompt, so an absent one is dropped rather than asserted as a literal `None`.

## Not included, on purpose

`LoopExit` / `exit_reason` — it lands with #1789, still open. This should gain it once that merges, rather than duplicating the enum here.

## Tests

21, including: extraction pairing an execute with the visualization it followed (not the last of each independently), a bare-URI field in raw tool-call arguments, the wrong horizon failing on period alone, an unstated expectation neither failing nor silently passing, a chat error keeping what its `partial_result` carried, and a chat error on a later run not discarding the earlier one.

801 passed, lint and format clean.

## Merge note

The two other PRs in this set touch the same three files — `cli/agentic_runner.py` plus the `_ALL_AGENTIC_KIND_CASES` and `_EVALUATE_FUNCS` staleness guards. Whichever merges first, the others need a trivial rebase on those lists.

## Summary by CodeRabbit

* **New Features**
  * Added support for agentic forecasting evaluations.
  * Forecasting runs can verify forecast activation, period selection, and metric accuracy across multiple attempts.
  * Added pass-rate summaries and detailed diagnostics for forecasting evaluation results.

* **Tests**
  * Added coverage for forecasting behavior, evaluation scoring, dispatching, and trace reporting.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
+++ b/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
@@ -12,6 +12,7 @@ from gooddata_eval.core.agentic._langfuse import make_langfuse_client
 from gooddata_eval.core.agentic._trace_linker import BackgroundTraceLinker, SubmitTraceLink, run_trace_link_inline
 from gooddata_eval.core.agentic.alert_skill import evaluate_agentic_alert_skill
 from gooddata_eval.core.agentic.conversation import ConversationFixture, evaluate_agentic_conversation
+from gooddata_eval.core.agentic.forecasting import evaluate_agentic_forecasting
 from gooddata_eval.core.agentic.general_question import evaluate_agentic_general_question
 from gooddata_eval.core.agentic.guardrail import evaluate_agentic_guardrail
 from gooddata_eval.core.agentic.kda_skill import evaluate_agentic_kda_skill
@@ -44,6 +45,7 @@ AGENTIC_TEST_KINDS = frozenset(
         "agentic_guardrail",
         "agentic_conversation",
         "agentic_kda_skill",
+        "agentic_forecasting",
     }
 )
 
@@ -219,6 +221,17 @@ def _dispatch_agentic(
         )
     elif kind == "agentic_kda_skill":
         return evaluate_agentic_kda_skill(
+            host=host,
+            token=token,
+            workspace_id=workspace_id,
+            question=item.question,
+            expected_output=eo if isinstance(eo, dict) else {},
+            k=k,
+            agent_id=agent_id,
+            **lf_kw,
+        )
+    elif kind == "agentic_forecasting":
+        return evaluate_agentic_forecasting(
             host=host,
             token=token,
             workspace_id=workspace_id,
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/forecasting.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/forecasting.py
@@ -526,8 +526,15 @@ def evaluate_agentic_forecasting(
                     ctx.quality(
                         tid,
                         strict_checks=strict_checks,
-                        latency_sec=run.turn_wall_clock_sec,
-                        cost_usd=pt.total_cost if pt and ev.triggered else None,
+                        # pt.latency covers the whole conversation, which is the item's real
+                        # elapsed cost when the agent needed clarification turns to get
+                        # there; turn_wall_clock_sec (the goal turn alone) is the fallback.
+                        # This is what 7 of the 8 existing kinds do -- kda_skill is the
+                        # outlier and documents its own reason. Cost is not gated on
+                        # ev.triggered: a run that answered without ever reaching the tool
+                        # still spent tokens, and hiding that understates what the item cost.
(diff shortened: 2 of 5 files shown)


Tomkess:
All three findings fixed in `59958ce0` (+ `273b8428` for a PERF403 follow-up). All three were real, and **two of them applied to the sibling PRs too** — fixed there as well, `#1799` in `ab988606` and `#1801` in `bed8974d`.

**Tool calls extracted from the current turn only** — the worst of the three, and it would have bitten the common path. The agent routinely asks which measure to forecast before building anything, so the chart lands on turn 2 and `execute_forecast` on turn 3; reading only the current turn dropped the visualization the forecast actually ran on, and the evaluator then failed a correct run for an empty config. Extraction now reads every turn accumulated so far.

Worth noting why `kda_skill` doesn't have this bug despite the identical structure: its create and execute always land in the same turn ("NO confirmation needed" in the skill's own prompt), so a single-turn read is sufficient there. I copied the shape without re-checking that assumption held for a skill that does ask questions first.

**`forecast_confidence` and `forecast_seasonal` unchecked** — correct, and it's the same failure as the unwired `max_widgets` you caught on #1797: I described them as checkable in the module docstring and the PR body, then never implemented them. Both are now scored. One detail: an absent `forecast_seasonal` counts as the tool's own default of `false` rather than a mismatch, so an agent that leaves the default alone isn't penalised.

**Unasserted checks scored as BOOLEAN 1** — right, and the reasoning is exactly as you put it. The internal `True` exists so an unstated expectation can't fail a run; publishing it as a score claims the evaluator verified something it never looked at. Only checks named in `ev.asserted` are scored now, on all three PRs.

5 tests added here (2 for confidence/seasonality, 1 for the tool default, 1 for the unasserted case, 1 for cross-turn extraction), 1 each on the siblings. The cross-turn test was verified to fail against the previous version on all three.

806 passed, lint and format clean.

hkad98 (dismissed):

Tomkess:
Thanks — the gate finding was real, and wider than one branch. Fixed in `0cf23ca3`.

`run_agentic_forecasting` already computed `pass_power_k`; the evaluator simply never read it, asking `if not summary.pass_at_k` for the verdict. So `--gate power` on a flaky item passed on the strength of one good run out of K. Forecasting now takes `gate` like the other multi-run kinds, uses `gate_passed(...)`, carries `gate_failure_note` into the assertion message (the message body describes the *best* run, which under pass^K can be one that passed), and publishes pass@K / pass^K / gate_passed to Langfuse.

Default behaviour is unchanged — without `--gate` it is pass@K exactly as before. Three tests cover it, and the power one fails against the previous code.

**Not fixed here, deliberately:** `agentic_what_if` (already on master) and `agentic_anomaly_detection` (#1801) have the same gap, and neither evaluator accepts `gate` at all. That is a follow-up covering both rather than a wider diff on this PR.

On the other three:

- **Pinned `forecast_confidence`/`forecast_seasonal` withheld from the clarification prompt** — agreed, and `is not None` is the right guard so a pinned `False` still reaches the simulated user. Worth doing.
- **Simulated-user failure scored as a forecast failure** — agreed it is wrong to attribute a harness fault to the agent. #1789 introduces `LoopExit.SIMULATED_USER_FAILED` for exactly this and applies it across the kinds; better handled there than duplicating the enum here.
- The comment at `forecasting.py:125` carries no finding text, only the static-analysis transcript, so there is nothing to action.

PR #1799, opened 2026-09-10
feat(gooddata-eval): add the agentic what-if-analysis evaluator

Tomkess:
Second of three evaluators for skills that ship in the product but have no eval coverage. Sibling PRs: #1798 (forecasting), anomaly detection to follow.

## Why now

Probed live against `micdiagnose-dev`: the skill **is** enabled and reachable — a what-if question activates `set_skills(["what_if_analysis", "visualization"])`. Nothing evaluates it.

## The most checkable of the three

The scenario spec carries adjustments shaped `{metric_id, metric_type, scenario_maql}`, and `scenario_maql` is the adjusted expression. From the tool's own description: a 10% uplift on a revenue metric defined as `SELECT SUM({fact/price} * {fact/quantity})` becomes `SELECT SUM({fact/price} * 1.10 * {fact/quantity})`.

That is MAQL, and MAQL already has a comparator in this package — `evaluators._maql.normalize_maql`, used by `metric_skill`. So "did it apply the right adjustment" is answerable exactly, without a judge.

| Check | Source |
|---|---|
| `triggered` / `executed` / `success` | the create/execute pair |
| `metric_correct` | `adjustments[].metric_id` |
| `maql_correct` | `adjustments[].scenario_maql`, normalized |
| `scenario_count_correct` | `len(scenarios)` |
| `baseline_correct` | `include_baseline` |

What the adjustment *produced* is deliberately not checked — that is the platform's arithmetic, not the agent's. Only that the agent asked for the right thing and the execution succeeded.

## Three details worth review

**A candidate list for MAQL.** `* 1.1` and `* 1.10` are the same uplift, and a rewrite can reach the same value, so `scenario_maql` accepts a string or a list — the same shape `metric_skill` uses for its accepted MAQLs.

**`include_baseline` defaults to true in the tool**, so an absent argument means the agent *did* ask for a baseline. Treating absent as false would fail correct behaviour; the check treats `None` as the default.

**An unstated expectation passes**, with `detail["asserted"]` recording which checks the fixture pinned — otherwise a run that verified nothing is indistinguishable from one where everything matched.

## The loop

Follows `kda_skill`. The agent asks which measure to adjust before building anything — observed live: *"I need to confirm which 'Spend' calculation you want to adjust (there are multiple in your data)."* A simulated user answers from the fixture's hints; absent hints are dropped rather than asserted as `None`.

## Not included, on purpose

`LoopExit` / `exit_reason` — lands with #1789, still open.

## Tests

22, including: extraction pairing an execute with the spec it followed, adjustments flattening across scenarios, MAQL compared normalized rather than literally, a candidate list accepting either uplift spelling, an absent `include_baseline` counting as the tool's default, and a malformed scenario list not raising.

802 passed, lint and format clean.

## Merge note

Touches the same three files as the sibling PRs — `cli/agentic_runner.py` plus the `_ALL_AGENTIC_KIND_CASES` and `_EVALUATE_FUNCS` staleness guards. Whichever merges first, the others need a trivial rebase on those lists.

## Summary by CodeRabbit

* **New Features**
  * Added support for agentic what-if analysis evaluations.
  * Evaluations can simulate clarification responses, assess scenario execution, and report pass-at-K results.
  * Added detailed validation for metrics, scenarios, MAQL, baselines, and completed conversations.
  * Integrated what-if evaluations with trace reporting.

* **Tests**
  * Added coverage for what-if execution, scoring, error handling, iteration limits, and evaluation dispatch.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
+++ b/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
@@ -18,6 +18,7 @@ from gooddata_eval.core.agentic.kda_skill import evaluate_agentic_kda_skill
 from gooddata_eval.core.agentic.metric_skill import evaluate_agentic_metric_skill
 from gooddata_eval.core.agentic.search_tool import evaluate_agentic_search_tool
 from gooddata_eval.core.agentic.visualization import evaluate_agentic_visualization
+from gooddata_eval.core.agentic.what_if import evaluate_agentic_what_if
 from gooddata_eval.core.config import ReasoningEffort
 from gooddata_eval.core.models import AgenticEvalOutcome, CreatedVisualization, DatasetItem
 from gooddata_eval.core.runner import EvalReport, ItemReport
@@ -44,6 +45,7 @@ AGENTIC_TEST_KINDS = frozenset(
         "agentic_guardrail",
         "agentic_conversation",
         "agentic_kda_skill",
+        "agentic_what_if",
     }
 )
 
@@ -219,6 +221,17 @@ def _dispatch_agentic(
         )
     elif kind == "agentic_kda_skill":
         return evaluate_agentic_kda_skill(
+            host=host,
+            token=token,
+            workspace_id=workspace_id,
+            question=item.question,
+            expected_output=eo if isinstance(eo, dict) else {},
+            k=k,
+            agent_id=agent_id,
+            **lf_kw,
+        )
+    elif kind == "agentic_what_if":
+        return evaluate_agentic_what_if(
             host=host,
             token=token,
             workspace_id=workspace_id,
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/what_if.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/what_if.py
@@ -339,13 +339,16 @@ def run_agentic_what_if(
                     reasoning_steps.extend(partial.reasoning_steps or [])
                     response_id = partial.response_id or response_id
                     _accumulate(partial)
-                    create_args, execute_result = _extract_what_if_calls(partial.tool_call_events or [])
+                    create_args, execute_result = _extract_what_if_calls(all_tool_call_events)
                 turn_completed = False
                 break
             reasoning_steps.extend(chat_result.reasoning_steps or [])
             response_id = chat_result.response_id or response_id
             _accumulate(chat_result)
-            create_args, execute_result = _extract_what_if_calls(chat_result.tool_call_events or [])
+            # Over every turn so far, not just this one: the agent may build the spec on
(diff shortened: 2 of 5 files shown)


Tomkess:
Review on the sibling PR #1798 surfaced two findings that are structural and applied here unchanged. Fixed in ``, before this PR was reviewed.

**Tool calls were extracted from the current turn only.** The agent asks a disambiguation question before building anything, so the create call and the execute call can land on different turns — reading a single turn dropped the object the execution actually ran on, and the evaluator then failed a correct run for having no content to check. Extraction now reads every turn accumulated so far. `kda_skill` does not have this bug despite the identical structure, because its create and execute always land in the same turn; I copied the shape without re-checking that assumption held for a skill that asks questions first.

**Unasserted content checks were published to Langfuse as BOOLEAN 1.** They are `True` internally so an unstated expectation cannot fail a run, but scoring that claims the evaluator verified something it never looked at. Only checks named in `ev.asserted` are scored now.

One test added, verified to fail against the previous version. Lint and format clean.

Tomkess:
Both fixed in `097d1912`, and the first one was the most valuable finding across this whole set.

**MAQL and metric matched independently.** Correct, and it was a false PASS, not a false FAIL: a wrong adjustment on the right measure plus a right adjustment on the wrong measure satisfied one check each, and the run passed with two actual failures. The MAQL check is now scoped to adjustments on the expected measure.

That also corrected a semantic I had wrong. `test_adjusting_the_wrong_measure_fails_on_metric_alone` asserted `maql_correct is True` when the wrong measure was adjusted — which was documenting the very independence that was the bug. With the pairing, `maql_correct` is now False there too, and that is the honest reading: the expected measure was not adjusted at all, correctly or otherwise. Test renamed and its docstring says why.

**Latency and cost.** Also right, and I checked the claim rather than taking it: 7 of the 8 existing kinds use `pt.latency if pt else None`, and `kda_skill` is the lone outlier with its own documented reason (*"pt can be any trace of the conversation, not necessarily the KDA turn"*). I had copied `kda_skill` without re-checking whether its reasoning transferred, and it doesn't — these kinds routinely spend clarification turns, so reporting only the goal turn understates the item's elapsed cost on exactly the runs where it matters most. Now `pt.latency if pt else run.turn_wall_clock_sec`, keeping the goal turn as a fallback rather than the majority's `None`.

The `ev.triggered` gate on cost was my own invention, also copied from `kda_skill`, and it's plainly wrong: a run that answered without ever reaching the tool still spent tokens, and reporting no cost understates the item. Gate removed.

Both latency/cost fixes applied to the siblings too — #1798 in `c7bef360`, #1801 in `36247ea4`. The MAQL pairing is what-if-specific.

2 tests added, the pairing one verified to fail against the previous version. 805 passed, lint and format clean.

hkad98 (approved):

PR #1801, opened 2026-09-10
feat(gooddata-eval): add the agentic anomaly-detection evaluator

Tomkess:
Third of three evaluators for skills that ship in the product but have no eval coverage. Siblings: #1798 (forecasting), #1799 (what-if analysis).

## Why now

Probed live against `micdiagnose-dev`: the skill **is** enabled and completes in a single turn — search, build a monthly-spend chart, `execute_anomaly_detection`. Nothing evaluates it.

## Honest about what can be checked

`execute_anomaly_detection` takes only `visualization_ref` and `max_points`. No threshold, no sensitivity, no expected anomaly — so unlike forecasting and what-if, **the tool call itself carries almost no assertable intent**.

What *is* assertable is the chart the detection ran on. The measure and the time granularity are the whole of "did it look at the right series", and getting either wrong makes the result meaningless however well the detection performed.

```json
{"metric": "metric/spend", "granularity": "MONTH"}
```

**The flagged-point count is reported but never asserted.** Whether a real series contains anomalies is a property of the data, not of the agent — a fixture demanding some would start failing the day the warehouse refreshed. The live probe returned `point_count: 0` and the agent correctly said so; that is a pass.

If the team would rather not add another kind whose signal is mostly "the chain ran" (the same limitation `kda_skill` carries, open in the eval repo since August), this is the one of the three to drop. Forecasting and what-if check real correctness; this one checks targeting.

## A correction, and a much narrower real issue

An earlier revision of this PR claimed gen-ai's granularity inference was ambiguous for `label/process_date.month`, because `date` and `month` both map to a granularity. **That was wrong, and the fault was in this module, not the service.** gen-ai's token map has no `date` key:

```python
_TOKEN_TO_GRANULARITY = {"hour": "HOUR", "day": "DAY", "week": "WEEK",
                         "week_us": "WEEK", "month": "MONTH",
                         "quarter": "QUARTER", "year": "YEAR"}
```

This module invented `"date": "DAY"` and dropped the service's `"hour"`. Fixed in `5a57f66c` — the map is now copied verbatim. Two consequences of the invented key, both real bugs in the evaluator:

| Reference | Service | This module, before |
|---|---|---|
| `label/process_date` | matches nothing → the tool **refuses** the call | scored `DAY` — grading a run the service never ran |
| `label/process_date.hour` | `HOUR` | unrecognised |

**What survives is narrower and genuine.** A reference naming *two* granularities does let set iteration order decide the service's answer — a snake_case `first_day_quarter.month` tokenizes to `{first, day, quarter, month}`, three of which map. The last-token rule stays for that reason: a scorer must not be a coin flip even where the thing it scores is one.

Scope, measured rather than asserted: **0 of 558 labels** in the eval workspace name two granularities. This workspace's `FirstDayQuarter.*` labels are camelCase, so they tokenize to one word and are safe; the same attribute named `first_day_quarter` would not be. So it is latent, not live, and I would not file it upstream on this evidence alone.

## Not included, on purpose

`LoopExit` / `exit_reason` — lands with #1789, still open.

## Tests

23, including: extraction pairing a detection with the chart it followed, the ambiguous-label case above, a field token winning over a disagreeing filter (matching the service's order), finding no anomalies still passing, and a chat error on a later run not discarding the earlier one.

803 passed, lint and format clean.

## Merge note

Touches the same three files as the siblings — `cli/agentic_runner.py` plus the `_ALL_AGENTIC_KIND_CASES` and `_EVALUATE_FUNCS` staleness guards. Whichever merges first, the others need a trivial rebase on those lists.

## Summary by CodeRabbit

* **New Features**
  * Added support for evaluating agentic anomaly-detection scenarios.
  * Supports clarification exchanges, visualization and detection-result validation, repeated runs, and pass-rate aggregation.
  * Reports execution performance, including latency and cost information.
  * Provides detailed failure information when anomaly-detection evaluations do not pass.
  * Supports partial results and robust handling of interrupted or unsuccessful agent interactions.

### DIFF
--- a/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
+++ b/packages/gooddata-eval/src/gooddata_eval/cli/agentic_runner.py
@@ -11,6 +11,7 @@ from typing import Any, TypedDict
 from gooddata_eval.core.agentic._langfuse import make_langfuse_client
 from gooddata_eval.core.agentic._trace_linker import BackgroundTraceLinker, SubmitTraceLink, run_trace_link_inline
 from gooddata_eval.core.agentic.alert_skill import evaluate_agentic_alert_skill
+from gooddata_eval.core.agentic.anomaly_detection import evaluate_agentic_anomaly_detection
 from gooddata_eval.core.agentic.conversation import ConversationFixture, evaluate_agentic_conversation
 from gooddata_eval.core.agentic.general_question import evaluate_agentic_general_question
 from gooddata_eval.core.agentic.guardrail import evaluate_agentic_guardrail
@@ -44,6 +45,7 @@ AGENTIC_TEST_KINDS = frozenset(
         "agentic_guardrail",
         "agentic_conversation",
         "agentic_kda_skill",
+        "agentic_anomaly_detection",
     }
 )
 
@@ -219,6 +221,17 @@ def _dispatch_agentic(
         )
     elif kind == "agentic_kda_skill":
         return evaluate_agentic_kda_skill(
+            host=host,
+            token=token,
+            workspace_id=workspace_id,
+            question=item.question,
+            expected_output=eo if isinstance(eo, dict) else {},
+            k=k,
+            agent_id=agent_id,
+            **lf_kw,
+        )
+    elif kind == "agentic_anomaly_detection":
+        return evaluate_agentic_anomaly_detection(
             host=host,
             token=token,
             workspace_id=workspace_id,
--- a/packages/gooddata-eval/src/gooddata_eval/core/agentic/anomaly_detection.py
+++ b/packages/gooddata-eval/src/gooddata_eval/core/agentic/anomaly_detection.py
@@ -380,13 +380,16 @@ def run_agentic_anomaly_detection(
                     reasoning_steps.extend(partial.reasoning_steps or [])
                     response_id = partial.response_id or response_id
                     _accumulate(partial)
-                    viz_args, execute_result = _extract_anomaly_calls(partial.tool_call_events or [])
+                    viz_args, execute_result = _extract_anomaly_calls(all_tool_call_events)
                 turn_completed = False
                 break
             reasoning_steps.extend(chat_result.reasoning_steps or [])
             response_id = chat_result.response_id or response_id
             _accumulate(chat_result)
-            viz_args, execute_result = _extract_anomaly_calls(chat_result.tool_call_events or [])
+            # Over every turn so far, not just this one: the agent may build the chart on
+            # one turn and detect on the next, and reading a single turn would drop the
+            # series the detection actually ran on.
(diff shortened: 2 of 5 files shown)


Tomkess:
Review on the sibling PR #1798 surfaced two findings that are structural and applied here unchanged. Fixed in ``, before this PR was reviewed.

**Tool calls were extracted from the current turn only.** The agent asks a disambiguation question before building anything, so the create call and the execute call can land on different turns — reading a single turn dropped the object the execution actually ran on, and the evaluator then failed a correct run for having no content to check. Extraction now reads every turn accumulated so far. `kda_skill` does not have this bug despite the identical structure, because its create and execute always land in the same turn; I copied the shape without re-checking that assumption held for a skill that asks questions first.

**Unasserted content checks were published to Langfuse as BOOLEAN 1.** They are `True` internally so an unstated expectation cannot fail a run, but scoring that claims the evaluator verified something it never looked at. Only checks named in `ev.asserted` are scored now.

One test added, verified to fail against the previous version. Lint and format clean.

Tomkess:
**Correction — the "service bug" I reported in this PR was not one, and the fault was mine.**

I claimed gen-ai's granularity inference was ambiguous for `label/process_date.month`, on the grounds that `date` → DAY and `month` → MONTH both match. The service's token map has no `date` key at all:

```python
_TOKEN_TO_GRANULARITY: dict[str, str] = {
    "hour": "HOUR", "day": "DAY", "week": "WEEK", "week_us": "WEEK",
    "month": "MONTH", "quarter": "QUARTER", "year": "YEAR",
}
```

**This module invented `"date": "DAY"`**, and separately dropped the service's `"hour"`. The ambiguity I described existed only in my own copy. Fixed in `5a57f66c`; the map is now verbatim, and the PR body is corrected.

The invented key was also two real bugs in the evaluator, not just a wrong claim:

- `label/process_date` — a date attribute with no granularity suffix. The service matches nothing and the tool **refuses the call** with "Could not infer anomaly granularity". This module scored `DAY`, so it would have graded granularity for a run the service never performed.
- `label/process_date.hour` — unrecognised here, `HOUR` in the service.

**What survives is real but much narrower.** A reference naming *two* granularities does leave the service's answer to set iteration order — a snake_case `first_day_quarter.month` tokenizes to `{first, day, quarter, month}`. The last-token rule stays for that reason.

I measured the scope rather than assuming it this time: **0 of 558 labels** in the eval workspace name two granularities. The `FirstDayQuarter.*` labels here are camelCase, so they tokenize to a single word and are safe — the same attribute named `first_day_quarter` would not be. Latent, not live, and not worth filing upstream on this evidence.

Worth stating plainly: I asserted a defect in someone else's service from a misread of my own code, and put it in a PR body and a commit message. The check that caught it was reading the service's map again instead of trusting my note about it.

805 passed, lint and format clean.

hkad98 (dismissed):

PR #1808, opened 2026-09-10
feat(gooddata-eval): generate eval datasets from insights, plus a readable HTML report

xvalovic:
**`gd-eval generate`** — builds a `visualization` dataset out of the charts a workspace
already has. Each insight's buckets/sorts/filters become the `expected_output`; an LLM
writes only the question, and any question that contradicts its own spec is rejected.
`--enrich-ranked N` additionally derives ranked items (real workspaces almost never
persist a sort, so that coverage was zero), preferring insights whose own title promised
a ranking their definition lacked; derived items carry `derived_from`/`derived_basis`.
`--skip-ambiguous` drops questions naming something the model carries more than once.

**`gd-eval report` / `run --html`** — self-contained HTML for one run or several side by
side: the whole conversation per item, each tool call with the arguments it got, and
per-step latency. `--redact` for customer-safe output.

## Summary by CodeRabbit

- **New Features**
  - Added `gd-eval report` for interactive, self-contained HTML reports with filtering, comparisons, titles, and optional redaction.
  - Added `gd-eval generate` to create visualization evaluation datasets from workspace insights, including ranking enrichment and ambiguity handling.
  - Added configurable turn and item timeouts for chat evaluations.
  - Added HTML output and redaction options to `gd-eval run`.
  - Added `any` and `power` evaluation gates with gate results in reports.

- **Improvements**
  - Evaluation details now include tool-call information alongside timeline data.
  - Improved date-granularity normalization for more consistent scoring.


packages/gooddata-eval/src/gooddata_eval/core/dataset/from_insights.py
```diff
@@ -0,0 +1,1372 @@
+# (C) 2026 GoodData Corporation
+"""Reverse-generate `visualization` dataset items from a workspace's real insights.
```
hkad98, line 2: This seems to be redundant. Cannot we use just logic from gooddata-sdk or gooddata-code-convertors?

xvalovic: You're right, there are some components in sdk that we can reuse, so SDK's reading now goes through get_declarative_analytics_model + get_declarative_ldm (see line  728)
I also took the granularity table from gooddata_sdk.visualization._GRANULARITY_CONVERSION, which fixed a real bug (month_in_year was becoming MONTH_IN_YEAR instead of MONTH_OF_YEAR).

packages/gooddata-eval/src/gooddata_eval/core/dataset/from_insights.py
```diff
@@ -0,0 +1,1387 @@
...
+    """The spec's sort entries."""
+    return spec["query"].get("sort_by") or []
+
+
+def _sort_field(sort: dict) -> str:
+    """The alias a sort entry orders on, whichever key its type puts it under."""
+    return sort["by"] if sort.get("type") == "attribute_sort" else sort["metrics"][0]
+
+
+def convert(viz: dict, date_instance_ids: set, display_names: dict | None = None) -> dict:
```
hkad98, line 389: This is not necessary. Please read README.md for https://pypi.org/project/gooddata-code-convertors/

Specifically: Available Converters

packages/gooddata-eval/src/gooddata_eval/core/reporting/report_template.html
```diff
@@ -0,0 +1,452 @@
+<!-- (C) 2026 GoodData Corporation -->
```
hkad98, line 1: Consider using JINJA template

xvalovic: I think jinja is not a option here. There are no loops, conditions or variables, that could favour jinja.
Morover, we would have to resolve:

- another dependency to package (jinja isn't a dep of gooddata-eval or gooddata-sdk)
- syntax conflicts between jinja and JS (e.g. {{ }} )
- replace str.replace() with jinja.render() - basically no benefit

In this phase, when we have only this single file (as more-or-less standalone SPA) without any complex statements, I would leave rendering solely to JS; Python only injects the JSON payload.

hkad98 (approved):

PR #1813, opened 2026-09-15
feat(gooddata-eval): record the turns and steps each agentic run took

myhoai:
The effort comparison in gdc-nas reads these. A pass/fail bit cannot separate two reasoning efforts, while the step count moves with the effort, and the turn count tells a wrong answer from a run max_iterations cut short.

alert_skill and kda_skill gained the counters; metric_skill and conversation already had them and only needed the write. conversation also logs clarification_turns, since its total includes the turns its fixture asks for.

## Summary by CodeRabbit

- **Bug Fixes**
  - Agentic alert, KDA, conversation, metric, and visualization evaluations now report turn and reasoning-step totals as accurate integer values.
  - Evaluation telemetry records integer turn and reasoning-step scores, including clarification turns where applicable.
  - Run metrics now correctly include clarification and creation turns and report zero activity when a run fails before producing a response.

- **Tests**
  - Added regression coverage for effort metrics, clarification handling, failed runs, and telemetry values across agentic evaluations.


packages/gooddata-eval/src/gooddata_eval/core/agentic/alert_skill.py
```diff
@@ -477,6 +477,8 @@ class AlertRunResult:
     alert_id: str | None
     eval: AlertEvaluation
     actual_alert_arguments: dict
+    total_turns: float = 0.0
+    total_steps: float = 0.0
```
hkad98: Nit: does this needs to be float?

hkad98 (approved):

